PR #41, opened 2026-05-28
Migrate Battery Tracker plugin (battery-tracker)

### LINKED ISSUE #9: Migrate plugin: Battery Tracker (battery-tracker)
> **Auto-detected** — source: `linux-gaming-plugin-manager/plugins/battery-tracker/` · UI: overlay (app.tsx) · removed-package deps to inline: none · subprocess: .

# Migrate one plugin: Battery Tracker (`battery-tracker`)

**Goal:** Port the `battery-tracker` plugin from the old Steam Loader repo into the new Loadout repo, one plugin at a time. Port it *faithfully* (zero behavior regressions), simplify only where it is provably safe, and keep all of its code *inside the plugin* unless something is genuinely shared by 2+ already-migrated plugins.

## Locations
  - This plugin: `plugins/battery-tracker/`
  - Port into: `plugins/battery-tracker/`

---

**Layout** — `plugins/battery-tracker/`:
- `package.json` — name `@loadout/plugin-battery-tracker`, `"type": "module"`, deps, and a `plugin` field (the manifest now lives here, NOT in a separate `plugin.json`):
  ```json
  {
    "name": "@loadout/plugin-battery-tracker",
    "version": "0.0.1",
    "type": "module",
    "plugin": {
      "id": "battery-tracker",
      "name": "Battery Tracker",
      "description": "…",
      "category": "…",
    }
  }
  ```

```bash
```

---

| Distro | Notes |
|---|---|

**Binary classification:**
  - `hyprctl` — Hyprland only.

---

## Porting procedure

1. **Copy** the source plugin tree `plugins/battery-tracker/` from SOURCE into TARGET `plugins/battery-tracker/`.
3. **Fold `plugin.json` into `package.json`.** Move the manifest fields into the `plugin` field of `package.json`, set `name` to `@loadout/plugin-battery-tracker`, add `"type": "module"` and the real `dependencies`. Delete the standalone `plugin.json`.

---

|---|---|---|
| `plugin-storage` | ×8 | Inline: read/write JSON at `~/.config/loadout/plugins/battery-tracker.json` (~30–50 LOC). |
| `sgdb-art` | ×2 | Inline. |

---

**Safe to cut / collapse:**

**NOT safe — do NOT touch:**

---

## Definition of Done

- [ ] `bun run typecheck`
- [ ] `bun run build`.
- [ ] **Plugin loads**: the loader logs `Loaded plugin: Battery Tracker (battery-tracker) …` (from `apps/loadout/src/loader/plugin-manager.ts`).

---

- **Plugin id:** `battery-tracker`
- **Plugin name:** `Battery Tracker`
- **Source path:** `linux-gaming-plugin-manager/plugins/battery-tracker/`

srsholmes:
Closes #9.

## Summary

- Ports `battery-tracker` from `linux-gaming-plugin-manager` into `loadout/plugins/battery-tracker/`
- Zero behavior regressions: same RPC methods (`getBatteryInfo`, `getHistory`), same event (`batteryUpdate`), same 10s update + 60s history cadence, same HID-peripheral scoring logic
- No subprocess — reads `/sys/class/power_supply/*` via `node:fs/promises` `readFile`/`readdir`; declared in `permissions.filesystem`
- `Bun.file` replaced by `readFile` throughout (eslint-compliant)
- Private methods renamed with `_` prefix to exclude them from the RPC surface
- `mountComponent` factory collapses `mount`/`mountHomeWidget`/`mountHeader` boilerplate

## Simplifications / extractions

- **`lib/battery.ts`** — pure sysfs parsing helpers extracted from backend: `parseSysfsInt`, `microToUnit`, `computeHealthPercent`, `computeTimeRemaining`, `formatTimeRemaining`, `assembleBatteryInfo`, `scoreBatteryCandidate`. 32 unit tests in `lib/battery.test.ts`.
- `mountComponent` factory collapses the identical `createRoot`/`PluginProvider`/`unmount` pattern shared by all three mount exports.
- Legacy migrations: none to delete (plugin had none).

## Removed-package deps

None — this plugin never used `plugin-storage`, `vdf`, or other removed packages.

## Test plan

- [x] `bun run typecheck` — clean
- [x] `bun run lint` — 0 errors
- [x] `bun run check:specs` — OK
- [x] `bun test plugins/battery-tracker/backend.test.ts` — 18 pass, 0 fail
- [x] `bun test plugins/battery-tracker/lib/battery.test.ts` — 32 pass, 0 fail
- [x] `bun test plugins/battery-tracker/app.spec.tsx --preload ./test/bun-test-setup.ts` — 14 pass, 0 fail
- [x] Full `bun test test.ts` — 10 pre-existing CDPClient failures confirmed absent in main and confirmed pre-existing before this commit (verified via `git stash`)

## Note on the inlined plugin-storage helper

This plugin does NOT use plugin-storage (no history persistence across restarts — history is an in-memory ring buffer). If a second migrated plugin needs an identical inlined storage helper, flag it for extraction then.


srsholmes:
## Automated migration review — PR #41 (battery-tracker)

### Code quality

- **Inlined `mountComponent` factory** (`app.tsx:433-447`): duplicates the shared `mountComponent` from `@loadout/ui` (`packages/ui/src/sdk.tsx:98`). The migration spec's DoD explicitly bans this: "Mount (UI): `mountComponent(Component)` + `mountHeaderStub` … from `@loadout/ui`. No inlined `createRoot + PluginProvider` boilerplate." Other migrated plugins (`bluetooth`, `tdp-control`, `fan-control`, `display-settings`, `playtime`, `steam-gamescope-ipc`) all import it. Drop local factory + `react-dom/client` import.
- **`BatteryInfoOrError` type leak** (`app.tsx:16`): `BatteryInfo & { error?: string }` exists only because `getBatteryInfo`'s RPC returns a union. Backend already types return as `BatteryInfo | { error: string }`; UI re-models as intersection then casts. Discriminated check (`"error" in info`) would avoid the cast.
- **`react-icons` undeclared in `package.json`** — imported at `app.tsx:3` but not in `dependencies`. Spec lists it as runtime-hoisted, so it works. `tdp-control` declares it explicitly — inconsistency.
- **Pure logic extraction exemplary**: `lib/battery.ts` holds `parseSysfsInt`, `microToUnit`, `round2`, `computeHealthPercent`, `computeTimeRemaining`, `formatTimeRemaining`, `assembleBatteryInfo`, `scoreBatteryCandidate`, all unit-tested. `backend.ts` is I/O + RPC plumbing only.
- **Audit D-018 collapsed timers cleanly** (`backend.ts:48-61`): single setInterval ticking history every 6th tick, faithful to source.

### Bundled-binary quality
No `bundled_bins` — pure sysfs reader. N/A.

### Lint-warnings count
- Baseline (main): **35** — PR: **37** — Delta: **+2**
- Both new warnings in `app.spec.tsx:22,25` (`(_method: string)` + `({ children }: any)` in `mock.module` override). Within "+2 per plugin" allowance.

### Test coverage
- `check:specs` pass.
- `lib/battery.test.ts`: 38 cases covering all 8 pure helpers. Strong.
- `backend.test.ts`: 14 cases covering `getBatteryInfo` (no battery / discharging / charging / Full / missing files / non-numeric / health calc / short format), `getHistory` (empty / accumulate / cap at 60 / null path), `_findBatteryPath` including the OXP Apex HID-scoring regression case.
- `app.spec.tsx`: 13 cases — mount, header, widget, percentage/power/health display, event subscription, error state, spinner, live update propagation.

| Permission | Class | Notes |
|---|---|---|
| `commands: []` | ✅ | No subprocesses. |
| `filesystem: /sys/class/power_supply` | ✅ | Kernel-provided everywhere. |

**Verdict:** ✅ likely on all five distros. Detect-and-degrade returns `{ error: "No battery detected" }` for desktops without batteries.

- **Power-draw sign is a UX lie, not a bug-bug** (`app.tsx:234-235`, `app.tsx:409`): sysfs `power_now` is a magnitude — always ≥ 0 — so `drawW >= 0` is always true. The "+" prefix and `--color-success` tile color appear even while discharging. **Source has the same bug**; faithful port preserved it. Worth a follow-up; the `charging` flag is the actual signal.
- **`Charging` status with `Full` formatted string** (`lib/battery.ts:97-99`): `formatTimeRemaining(0, "Charging")` returns `"0m"` but `formatTimeRemaining(null, "Charging")` returns `"--"`. The deficit-zero path at `lib/battery.ts:85-87` only fires when `powerWatts > 0`. If `Charging` reported with `powerWatts=0` (happens transiently on adapter unplug→replug), user sees `"--"` until next 10s tick.
- **`mock.module("@loadout/ui", …)` in `app.spec.tsx:18`** is correct for bun's non-hoisted mock, but the `mockHomeWidget` test never asserts the body class — only that an unmount function is returned. Light coverage on home-widget path; no live-update assertion against it.

---

srsholmes:
Tested on Steam Deck (SteamOS 3.7) and Apex — backend loads cleanly, battery candidate scoring correctly picks the system battery over HID gamepad batteries (`BAT1(15)` vs `hid-…-battery(-140)`), and the frontend bundle compiles (50 modules, 1.0 MB). Overlay panel renders and updates.

PR #56, opened 2026-05-28
Migrate InputPlumber plugin (input-plumber)

### LINKED ISSUE #15: Migrate plugin: InputPlumber (input-plumber)
> **Auto-detected** — source: `linux-gaming-plugin-manager/plugins/input-plumber/` · UI: overlay (app.tsx) · removed-package deps to inline: none · subprocess: .

# Migrate one plugin: InputPlumber (`input-plumber`)

**Goal:** Port the `input-plumber` plugin from the old Steam Loader repo into the new Loadout repo, one plugin at a time. Port it *faithfully* (zero behavior regressions), simplify only where it is provably safe, and keep all of its code *inside the plugin* unless something is genuinely shared by 2+ already-migrated plugins.

## Locations
  - This plugin: `plugins/input-plumber/`
  - Port into: `plugins/input-plumber/`

---

**Layout** — `plugins/input-plumber/`:
- `package.json` — name `@loadout/plugin-input-plumber`, `"type": "module"`, deps, and a `plugin` field (the manifest now lives here, NOT in a separate `plugin.json`):
  ```json
  {
    "name": "@loadout/plugin-input-plumber",
    "version": "0.0.1",
    "type": "module",
    "plugin": {
      "id": "input-plumber",
      "name": "InputPlumber",
      "description": "…",
      "category": "…",
    }
  }
  ```

```bash
```

---

| Distro | Notes |
|---|---|

**Binary classification:**
  - `hyprctl` — Hyprland only.

---

## Porting procedure

1. **Copy** the source plugin tree `plugins/input-plumber/` from SOURCE into TARGET `plugins/input-plumber/`.
3. **Fold `plugin.json` into `package.json`.** Move the manifest fields into the `plugin` field of `package.json`, set `name` to `@loadout/plugin-input-plumber`, add `"type": "module"` and the real `dependencies`. Delete the standalone `plugin.json`.

---

|---|---|---|
| `plugin-storage` | ×8 | Inline: read/write JSON at `~/.config/loadout/plugins/input-plumber.json` (~30–50 LOC). |
| `sgdb-art` | ×2 | Inline. |

---

**Safe to cut / collapse:**

**NOT safe — do NOT touch:**

---

## Definition of Done

- [ ] `bun run typecheck`
- [ ] `bun run build`.
- [ ] **Plugin loads**: the loader logs `Loaded plugin: InputPlumber (input-plumber) …` (from `apps/loadout/src/loader/plugin-manager.ts`).

---

- **Plugin id:** `input-plumber`
- **Plugin name:** `InputPlumber`
- **Source path:** `linux-gaming-plugin-manager/plugins/input-plumber/`

srsholmes:
## Summary

Faithful port of the steam-loader `input-plumber` plugin into loadout. The plugin
installs the InputPlumber input-routing daemon on distros that don't ship it
(CachyOS, mutable Fedora, atomic Fedora variants without IP in the base image).

- **backend.ts**: thin orchestrator over `lib/install.ts` — 30s status broadcast,
  serialised install runs framed with `install-state` + `install-log` events,
  cancellation handle, status caching (60s TTL).
- **lib/install.ts**: probe + script runner. Replaces the source's
  `src/privileged.ts` (the entire sudo wrapper module is gone — the loadout
  backend runs as root). Calls into `@loadout/exec` (`runFull` + `spawn`) directly.
- **app.tsx**: install card with chip + buttons + live log pane. Uses
  `mountComponent(Panel) + mountHeaderStub` — dropped inline `createRoot` +
  `PluginProvider` boilerplate.
- **scripts/install-inputplumber.sh**: dropped the `id -u 0` self-check (backend
  is root); otherwise unchanged from source.

Closes #15.

## Per-distro verdict

`permissions.commands` and `permissions.filesystem` audit:

| Command | Classification | Notes |
|---|---|---|
| `bash`, `which`, `systemctl`, `curl`, `tar`, `sha256sum`, `chmod`, `cp`, `mkdir`, `rm`, `tee` | Universal | base coreutils + systemd; present on all five distros. |
| `ldconfig`, `udevadm` | Universal | glibc + systemd-udev; present everywhere. |
| `chcon` | Session-conditional | SELinux toolchain; present on Fedora/Bazzite/Nobara, may be absent on stock Arch/SteamOS. Script no-ops via `\|\| true`. |
| `pacman` | Distro-specific | Arch/CachyOS/SteamOS/ChimeraOS only. Script gates by `command -v pacman`. |
| `dnf` | Distro-specific | Fedora/Bazzite/Nobara only. Script gates by `command -v dnf`. |
| `inputplumber` | Bundled-or-installed | The point of the plugin: install if missing. |

`permissions.filesystem` paths are all standard `/etc` system-config locations
plus `/var/lib/inputplumber` for our own install. Universal.

`permissions.network`: `github.com`, `api.github.com`, `objects.githubusercontent.com`
— declared for transparency. These are fetched by the bundled bash script via
`curl` in a subprocess, not in-process `fetch()`, so the sandboxed-fetch list
doesn't actively gate them; they're listed so reviewers can see what the script
calls.

**Distros: SteamOS OK | CachyOS OK | Bazzite OK | Nobara OK | ChimeraOS OK**

All five paths work because the install script branches:
- Arch-family with pacman → `pacman -S inputplumber` (works on SteamOS, CachyOS,
  ChimeraOS — InputPlumber is in Arch repos).
- Mutable Fedora with dnf → `dnf install inputplumber` (works on Nobara, vanilla
  Fedora; warned-and-falls-through if the COPR isn't enabled).
- `rpm-ostree`-booted (Bazzite) → tarball fallback to `/var/lib/inputplumber`,
  ostree-safe (no `/usr` writes). Synthesises a systemd unit pointing into `/var`.

Polkit `.policy` is best-effort onto `/usr/share/polkit-1/actions/` (read-only
on locked ostree deployments — script warns and continues; the daemon's own
dbus policy in `/etc` still works for most clients). This matches the source
plugin's behaviour and the project memory's `project_inputplumber_polkit_install.md`
notes.

## Test plan

- [x] `bun run typecheck`
- [x] `bun run lint` — 35 warnings (baseline), 0 errors, no regression
- [x] `bun run check:specs` — OK
- [x] `bun run test:backend` — 764 pass / 0 fail (21 new from this plugin)
- [x] `bun run test:ui` — 159 pass / 0 fail (6 new from this plugin)
- [x] `bun run build` — OK


srsholmes:
## Automated migration review — PR #56 (input-plumber)

### Code quality
- **Type duplication**: `app.tsx:24-35` redefines `ManagedBy`, `InstallStatus`, `InstallRunResult`, `InstallLogEvent`, `InstallStateEvent` inline. Backend exports `InstallStartResult` and re-uses `installer.InstallStatus`/`installer.InstallRunResult` from `lib/install.ts`. Comment even says "mirror backend.ts" — drifts the moment backend changes. Should share via a `types.ts`.
- **Console noise in backend** (`backend.ts:58, 63, 71, 111`): four `console.*` lines on lifecycle. Other migrated plugins use `this.log?.` — inconsistent.
- **`statusInflight` race**: cleared inside `try`/`finally` (`backend.ts:87-95`); second concurrent caller can see `statusInflight === null` and re-enter. TTL cache catches it on the second pass.
- **Polkit memory cross-check passes** — the 2026-05-03 polkit fix is faithfully ported (`.policy` install at lines 216-223, `.rules` install at lines 209-214).
- **`_invalidateStatusCache` underscore prefix**: source called it `invalidateStatusCache` (public-ish, exposed as RPC). Port renames to keep it off the wire — intentional, behavior-preserving.

### Bundled-binary quality
No `bundled_bins` declared. Script downloads upstream tarball at install time. BSD-3 + GPL-3 license separation explained in `NOTICE`. Network domains declared.

### Lint-warnings count
Baseline (main): 35 — PR: 35 — Delta: 0. No new errors. Zero new `any` casts in plugin code.

### Test coverage
- `check:specs`: pass · `bun test plugins/input-plumber`: 21 backend + 6 UI pass (27/27) · typecheck pass · build pass.

| Path | Class | Notes |
|---|---|---|
| `bash`, `which`, `tee`, `cat`, `cp`, `mkdir`, `rm`, `chmod`, `tar`, `curl`, `sha256sum` | ✅ | Universal coreutils. |
| `systemctl`, `udevadm`, `ldconfig` | ✅ | Universal on systemd distros. |
| `chcon` | ⚠️ | SELinux tool; gracefully `|| true`'d. |
| `pacman` / `dnf` | ⚠️ | Gated by `command -v` + `ID_LIKE`. |
| `/usr/share/polkit-1/actions` | ⚠️ | **Read-only on Bazzite/SteamOS** unless `/usr` unlocked. Script warns — InputPlumber clients then fail with "Action not registered" (matches `project_inputplumber_polkit_install.md` memory). |

**Verdict (out of the box):**
- **SteamOS (stock Deck): ⚠️ partial** — pacman fast path works on Arch base; polkit `.policy` write fails on immutable `/usr`. dbus policy survives in `/etc`. Clients get "Action not registered" on `busctl get-property`.
- **CachyOS: ✅ likely** — pacman fast path, writable `/usr`.
- **Bazzite: ⚠️ partial** — `IS_RPM_OSTREE=1` short-circuits to tarball; polkit `.policy` write under rpm-ostree warns only. Same caveat as SteamOS.
- **Nobara: ✅ likely** — dnf fast path.
- **ChimeraOS: ✅ likely** — Arch-based pacman path.

- `statusInflight` race window (above).
- `logs` slice memory in `app.tsx:120-126`: unbounded array.push + `.slice` allocates new array each chunk — chatty installs thrash GC.

---

PR #65, opened 2026-05-28
Migrate HowLongToBeat plugin (hltb)

### LINKED ISSUE #28: Migrate plugin: HLTB (hltb)
> **Auto-detected** — source: `linux-gaming-plugin-manager/plugins/hltb/` · UI: overlay (app.tsx) · removed-package deps to inline: external-cache,vdf · subprocess: .

# Migrate one plugin: HLTB (`hltb`)

**Goal:** Port the `hltb` plugin from the old Steam Loader repo into the new Loadout repo, one plugin at a time. Port it *faithfully* (zero behavior regressions), simplify only where it is provably safe, and keep all of its code *inside the plugin* unless something is genuinely shared by 2+ already-migrated plugins.

## Locations
  - This plugin: `plugins/hltb/`
  - Port into: `plugins/hltb/`

---

**Layout** — `plugins/hltb/`:
- `package.json` — name `@loadout/plugin-hltb`, `"type": "module"`, deps, and a `plugin` field (the manifest now lives here, NOT in a separate `plugin.json`):
  ```json
  {
    "name": "@loadout/plugin-hltb",
    "version": "0.0.1",
    "type": "module",
    "plugin": {
      "id": "hltb",
      "name": "HLTB",
      "description": "…",
      "category": "…",
    }
  }
  ```

```bash
```

---

| Distro | Notes |
|---|---|

**Binary classification:**
  - `hyprctl` — Hyprland only.

---

## Porting procedure

1. **Copy** the source plugin tree `plugins/hltb/` from SOURCE into TARGET `plugins/hltb/`.
3. **Fold `plugin.json` into `package.json`.** Move the manifest fields into the `plugin` field of `package.json`, set `name` to `@loadout/plugin-hltb`, add `"type": "module"` and the real `dependencies`. Delete the standalone `plugin.json`.

---

|---|---|---|
| `plugin-storage` | ×8 | Inline: read/write JSON at `~/.config/loadout/plugins/hltb.json` (~30–50 LOC). |
| `sgdb-art` | ×2 | Inline. |

---

**Safe to cut / collapse:**

**NOT safe — do NOT touch:**

---

## Definition of Done

- [ ] `bun run typecheck`
- [ ] `bun run build`.
- [ ] **Plugin loads**: the loader logs `Loaded plugin: HLTB (hltb) …` (from `apps/loadout/src/loader/plugin-manager.ts`).

---

- **Plugin id:** `hltb`
- **Plugin name:** `HLTB`
- **Source path:** `linux-gaming-plugin-manager/plugins/hltb/`

srsholmes:
## Summary

Migrates HowLongToBeat (HLTB) plugin from steam-loader. Closes #28.

## What's in this PR

- Plugin scaffold: `package.json` with `plugin.*` manifest, `permissions.network` for `howlongtobeat.com` + `store.steampowered.com`.
- Backend (`backend.ts`): uses `@loadout/{steam-cdp, external-cache, vdf}`. Manages multiple concurrent CDP connections (SharedJSContext + BPM + every `MainMenu_uid<N>` popup + store tabs) — uses raw `CDPClient` rather than `withSteamClient` (single-session) for that reason.
- Overlay UI (`app.tsx`): card grid of installed games + HLTB time badges. Detail view on click.
- **Consumes `__core:game-library`** for the library/collections list (single source of truth across SGDB / LSFG-VK / ProtonDB / here). Was previously coupled to the deprecated game-browser plugin.
- `GameInfo` / `GameCollection` imported from `@loadout/types` (was previously a local mirror).

## Deviations from source

- Settings + cache dir paths renamed from `steam-loader` → `loadout` (no back-compat).
- Tests converted from vitest → bun:test.

## CI gates

All green: typecheck, lint (35 warnings baseline), check:specs, test:backend, test:ui, build.


srsholmes:
## PR #65 — HowLongToBeat migration review

### Code quality
- **Mount boilerplate** (`app.tsx:1201-1228`): inlines `createRoot + PluginProvider` for `mount` / `mountHomeWidget` instead of `mountComponent(...)` from `@loadout/ui`. Every other migrated plugin (bluetooth, playtime, fan-control, tdp-control, steam-gamescope-ipc) uses `mountComponent` / `mountHeaderStub` — convention slip noted in the migration prompt's "Mount" rule.
- **Settings persistence convention slip** (`backend.ts:1681-1706`): writes to `~/.config/loadout/hltb/settings.json` via `Bun.file` / `Bun.write`. `@loadout/plugin-storage` (`readPluginStorage` / `writePluginStorage`) is the documented surface and lands at `~/.config/loadout/plugins/hltb.json`. The plugin imports neither `@loadout/plugin-storage` nor declares it as a dep. Also: the comment on `dataDir` (line 188-192) calls out an old test-suite footgun that suggests this was lifted from the legacy path rather than rewritten on the new convention.
- **`shortcuts.vdf` parse duplication** (`backend.ts:412-458`): re-reads userdata + parses binary VDF locally. `@loadout/game-library` already does this and the comment acknowledges it. Inline-until-2 rule applies, so flag-only — this is the **first** duplicate, no extraction yet.
- **Pure functions are already small** (`formatTime`, `normalize`, `lookupTimesByName`'s scoring tail) — all module-scope, no I/O. They're covered by `backend.test.ts` cases, no `lib/` extraction needed at this size.
- No `@steam-loader/*` remnants, no `sudo`/`pkexec`, no raw `Bun.spawn`. CDP client comes from `@loadout/steam-cdp` (workspace package), which the migration spec explicitly sanctions for Steam-CEF drivers. No cross-plugin imports.

### Bundled-binary quality
No `bundled_bins`. Pure network + Steam CEF integration.

### Lint-warnings count
| Source | Count |
|---|---|
| Baseline (`main` @ 82bb6e1) | 35 |
| PR (`plugin/hltb` @ 0489fba) | **37** |
| Delta | **+2** |

Both new warnings are in `app.spec.tsx:24` and `app.spec.tsx:31` — `: any` on the `useEvent` / `PluginProvider` mock shim signatures. 0 errors. 0 `as any` casts in `backend.ts` / `app.tsx` / `backend.test.ts`.

### Test coverage
- `bun run check:specs`: ✅ pass.
- **Backend**: 65 `it()` cases across 15 `describe()` blocks (`backend.test.ts` — 1664 LOC). Covers search, getTimes pipeline, profile_steam vs name vs fuzzy matching, auth retry on 403/404, cache hydration + clear, `fetchNextJsBuildKey` edge cases, and `pushBadgeDataToBPM` fan-out + coalescing.
- **UI**: 11 specs in `app.spec.tsx` (433 LOC) — settings save, library grid render, detail view, config card, home widget paths.
- Style: `mock.module("@loadout/ui", ...)` in `app.spec.tsx`. Skill spec allows this when SUT uses destructured / JSX imports (which is the case for `<PluginProvider>` / `useBackend`), and `--isolate` neutralises the cross-file leak risk. Fine as-is.

|---|---|---|
| `howlongtobeat.com` (network) | ✅ Universal | Vanilla HTTPS. |
| `store.steampowered.com` (network) | ✅ Universal | Used for the appdetails fallback. |
| `localhost:8080` (Steam CEF DevTools) | ⚠️ Session-conditional | Requires Steam launched with `-cef-enable-debugging`. Same dependency as every other CEF-driving plugin. |
| `~/.config/loadout/hltb/` (filesystem) | ✅ Universal | XDG-compatible — but **non-standard path** (see settings convention slip above). |
| `~/.cache/loadout/hltb/` (filesystem) | ✅ Universal | Honours `XDG_CACHE_HOME` via `@loadout/external-cache`. |
| `userdata/<uid>/config/shortcuts.vdf` (read) | ✅ Universal | Standard Steam layout. |
| `appmanifest_*.acf` (via `@loadout/steam-paths`) | ✅ Universal | Same package SGDB/LSFG-VK use. |

**Verdict (out of the box):**
- SteamOS (stock Deck): ✅ likely — Gaming Mode launches Steam with CEF debugging.
- CachyOS: ⚠️ partial — works in BPM; desktop Steam needs the user to opt into `-cef-enable-debugging`.
- Bazzite: ⚠️ partial — same as CachyOS for desktop sessions; Gaming Mode ✅.
- Nobara: ⚠️ partial — same caveat (desktop session needs the CEF flag).
- ChimeraOS: ✅ likely — handheld Gaming Mode mirrors SteamOS.

- **Fuzzy-match silent mis-identification** (`backend.ts:806-813`): when HLTB returns results but `fuzzysort.go` returns zero hits, the code falls back to the highest-`comp_all_count` (most-popular) result. For an obscure shortcut/emulator title with no real HLTB match, the badge will confidently render the wrong game's times. The 12h cache then locks the wrong match in. Worth at least negative-result caching or a min-score threshold.
- **Duplicate-title store tab leak** (`backend.ts:1303-1312`): `connections.set(\`store:\${tab.title}\`, conn)` overwrites without closing the prior `conn.client` if two store tabs share a title. WS socket leaks until process exit.
- **`getTimesForGame` `name` arg is not part of the cache key** (`backend.ts:724-754`): if a user renames a non-Steam shortcut, the next lookup returns the stale match cached against the old name because the cache is keyed on `appId` alone. Disk cache amplifies it for 12h.
- **`updateSettings` re-injects on every settings change** (`backend.ts:1077`): a rapid sequence of toggle clicks fans out a fresh `injectBadgeSystem()` per click — each iterates `bpmRenderKeys` + the store-tab loop with full CSS + script re-injection. No debounce.
- **Re-entrancy guard on `pushBadgeDataToBPM` only**: `injectBadgeSystem` and `pollCurrentAppId` can race each other via `checkHealth` triggering a `tryConnect` + re-inject while a poll is mid-flight (`polling` guard sits on its own state). Probably benign but the connections-map mutation inside `tryConnect` is not coordinated with the poller iterating `connections`.

---

PR #66, opened 2026-05-28
Migrate Launch Options plugin (launch-options)

### LINKED ISSUE #17: Migrate plugin: Launch Options (launch-options)
> **Auto-detected** — source: `linux-gaming-plugin-manager/plugins/launch-options/` · UI: overlay (app.tsx) · removed-package deps to inline: vdf · subprocess: .

# Migrate one plugin: Launch Options (`launch-options`)

**Goal:** Port the `launch-options` plugin from the old Steam Loader repo into the new Loadout repo, one plugin at a time. Port it *faithfully* (zero behavior regressions), simplify only where it is provably safe, and keep all of its code *inside the plugin* unless something is genuinely shared by 2+ already-migrated plugins.

## Locations
  - This plugin: `plugins/launch-options/`
  - Port into: `plugins/launch-options/`

---

**Layout** — `plugins/launch-options/`:
- `package.json` — name `@loadout/plugin-launch-options`, `"type": "module"`, deps, and a `plugin` field (the manifest now lives here, NOT in a separate `plugin.json`):
  ```json
  {
    "name": "@loadout/plugin-launch-options",
    "version": "0.0.1",
    "type": "module",
    "plugin": {
      "id": "launch-options",
      "name": "Launch Options",
      "description": "…",
      "category": "…",
    }
  }
  ```

```bash
```

---

| Distro | Notes |
|---|---|

**Binary classification:**
  - `hyprctl` — Hyprland only.

---

## Porting procedure

1. **Copy** the source plugin tree `plugins/launch-options/` from SOURCE into TARGET `plugins/launch-options/`.
3. **Fold `plugin.json` into `package.json`.** Move the manifest fields into the `plugin` field of `package.json`, set `name` to `@loadout/plugin-launch-options`, add `"type": "module"` and the real `dependencies`. Delete the standalone `plugin.json`.

---

|---|---|---|
| `plugin-storage` | ×8 | Inline: read/write JSON at `~/.config/loadout/plugins/launch-options.json` (~30–50 LOC). |
| `sgdb-art` | ×2 | Inline. |

---

**Safe to cut / collapse:**

**NOT safe — do NOT touch:**

---

## Definition of Done

- [ ] `bun run typecheck`
- [ ] `bun run build`.
- [ ] **Plugin loads**: the loader logs `Loaded plugin: Launch Options (launch-options) …` (from `apps/loadout/src/loader/plugin-manager.ts`).

---

- **Plugin id:** `launch-options`
- **Plugin name:** `Launch Options`
- **Source path:** `linux-gaming-plugin-manager/plugins/launch-options/`

srsholmes:
## Summary

Migrates Launch Options plugin from steam-loader. Closes #17.

## What's in this PR

- Plugin scaffold with `plugin.*` manifest, `permissions.filesystem` declarations.
- Backend (`backend.ts`): uses `@loadout/{steam-cdp, vdf}` to write Steam launch options for installed games via the running JS client (`SetLaunchOptionsForApp`).
- Per-game persistence at `~/.config/loadout/launch-presets.json`.
- Overlay UI (`app.tsx`): card grid of installed games + per-game launch-option editor.
- **Consumes `__core:game-library`** for the library/collections list. Was previously coupled to the deprecated game-browser plugin.
- `GameInfo` imported from `@loadout/types` (was previously a local mirror).

## Deviations from source

- Removed unnecessary permissions (`system: [\"exec\"]` was wrong — no subprocesses; `network: [\"localhost\"]` was wrong — `@loadout/steam-cdp` uses raw WebSocket, not sandboxed-fetch).
- Config path renamed `steam-loader` → `loadout` (no back-compat).
- Tests converted from vitest → bun:test.

## CI gates

All green: typecheck, lint (35 warnings baseline), check:specs, test:backend, test:ui, build.


srsholmes:
## PR #66 — launch-options migration review

### Code quality

- **Convention slip — mount boilerplate is inlined** (`app.tsx:780-803`). Other migrated plugins (`playtime`, `bluetooth`, `display-settings`) use `mountComponent(LaunchOptionsManager)` + `mountHeaderStub` from `@loadout/ui`. The hand-rolled `createRoot + PluginProvider` + no-op `mountHeader` is exactly what those helpers exist to replace. `PluginMountOpts` already includes `headerSlot`, so the factory handles this case.
- **Convention slip — preset persistence inlines fs instead of `@loadout/plugin-storage`** (`backend.ts:28-33`, `460-513`). The plugin ships its own `PRESETS_PATH = ~/.config/loadout/launch-presets.json` + `readFile/JSON.parse/mkdir/writeFile` quartet across `getPresets`/`savePreset`/`deletePreset`. Canonical shape is `readPluginStorage<{ presets: Record<string,string> }>("launch-options")` (see `plugins/playtime/backend.ts:238,272`). The migration prompt calls plugin-storage out by name as the always-use path.
- **Dead-ish import** (`app.tsx:13`): `export { FaTerminal as icon }` is fine, but `FaTerminal` is also imported at line 8 and used at 574. The re-export form is unusual — every other migrated plugin uses `export const icon = FaIcon;`. Style nit only.
- **Loosely typed `as unknown as` in spec mock** (`app.spec.tsx:21`). Acceptable in test glue per the migration prompt's "+2 `as any` per plugin" allowance, but worth noting.
- No `@steam-loader/*` remnants, no `sudo`/`pkexec`, no `Bun.spawn`, no cross-plugin imports. Imports are clean. The E-006 write-lock (`writeLock` + `withVdfLock`, `backend.ts:194-206`) is well-commented and the concurrent test (`backend.test.ts:262-353`) exercises both serialisation cases.
- Decomposition: `backend.ts` is 514 LOC but I/O-bound; only `getAppsSection` (`backend.ts:164-171`) is genuinely pure and it's 8 lines — too small to move to `lib/` on its own. `app.tsx` at 803 LOC could split `LaunchOptionsCard` (`:702-776`) into its own file, but it's already a clean sub-component.

### Bundled-binary quality

No `bundled_bins` — plugin is fs + CDP only.

### Lint-warnings count

Baseline (main): **35** — PR: **35** — Delta: **0**. 0 errors. No new `as any` in plugin source files.

- `bun run check:specs`: **pass** ("every plugin entry point ≥ 100 LOC has a sibling spec").
- `bun run test:backend`: **922 pass / 0 fail** including the two E-006 concurrent-write specs.
- backend.ts pure helpers worth extracting: `getAppsSection` is the only candidate and it's too short (8 LOC) to justify a separate `lib/`. `findLocalConfigs` and `readAllShortcutLaunchOptions` both do I/O — keep inline.
- `app.spec.tsx` covers mount/header portal/grid render/Configured pill/click-to-detail/back nav/presets view (9 specs). No coverage of the preset CRUD interactions in the presets view body (Save/Delete buttons), but those go through `backend.test.ts` already.

| Resource | Classification | Notes |
|---|---|---|
| `~/.config/loadout/launch-presets.json` (fs) | ✅ Universal | Plain home-dir JSON. |
| `~/.local/share/Steam/userdata` (fs) | ✅ Universal | Standard Steam path on all five distros. |
| `localconfig.vdf` parse + patch | ✅ Universal | Pure file I/O via `@loadout/vdf`. |
| `shortcuts.vdf` binary parse | ✅ Universal | Only the file format differs by Steam version, not by distro. |
| `withSteamClient` → `http://localhost:8080/json` | ⚠️ Session-conditional | Requires Steam running with `-cef-enable-debugging` (loader injects this). When Steam isn't reachable, backend falls back to direct VDF write — but see "Potential bugs" below: the loader's sandbox **will block this fetch** because no `permissions.network` is declared. |
| `permissions.commands` | ✅ none declared | No subprocesses — clean. |

**Verdict (out of the box):**
- SteamOS (stock Deck): ⚠️ partial — preset CRUD + VDF fallback work; Steam-Client live-update path is blocked by missing `permissions.network`.
- CachyOS: ⚠️ partial — same as above.
- Bazzite: ⚠️ partial — same as above.
- Nobara: ⚠️ partial — same as above.
- ChimeraOS: ⚠️ partial — same as above.

(All five degrade the same way: the VDF-fallback path keeps the plugin functional, but the live-update promise in the `setLaunchOptions` docstring is silently broken — Steam will overwrite the file on its next save and the user has to restart Steam to see changes.)

- **Missing `permissions.network`** (`package.json:20-25`). `withSteamClient` (`backend.ts:94`) goes through `fetch("http://localhost:8080/json")` in `@loadout/steam-cdp/tabs.ts:73`. The loader's `createSandboxedFetch` (`apps/loadout/src/loader/sandboxed-fetch.ts:32-46`) is deny-by-default and rejects every fetch when `permissions.network` is empty/missing, even for loopback (loopback is only auto-allowed when `localhost`/`127.0.0.1`/`::1` is *explicitly* listed). Net effect: `trySetViaSteamClient` will always throw, the warn-log on `backend.ts:97-100` will fire on every save, and the plugin will always take the VDF-write fallback. The plugin still "works" because the fallback exists, but the live-update behaviour the docstring promises (`backend.ts:298-310`) is silently broken — Steam will then clobber the VDF on its next autosave cycle.
- **`headerSlot` prop on `PluginProvider` may not match the real SDK shape** (`app.tsx:786-791`). The spec stubs out `PluginProvider` and wires `headerSlot` to `PluginHeaderSlotProvider` manually (`app.spec.tsx:21-43`). The real `mountComponent` (`packages/ui/src/sdk.tsx:101-112`) does pass `headerSlot` through, so this is probably fine — but worth a runtime smoke test that the dynamic header actually portals into the topbar slot on a real overlay. The fact that the spec has to reimplement `PluginProvider` to test header portal-ing is a smell.
- **`refresh()` swallows `getGames` errors silently** (`app.tsx:115-119`). Each of the four `Promise.all` legs `.catch(() => [])`. If `__core:game-library::getGames` fails, the user sees an empty library with no surface diagnostic; the outer try/catch on `:134-136` only catches what slips through. Not a regression vs. the source, just noted.

PR #71, opened 2026-05-28
Migrate Storage Cleaner plugin (storage-cleaner)

### LINKED ISSUE #10: Migrate plugin: Storage Cleaner (storage-cleaner)
> **Auto-detected** — source: `linux-gaming-plugin-manager/plugins/storage-cleaner/` · UI: overlay (app.tsx) · removed-package deps to inline: plugin-storage · subprocess: .

# Migrate one plugin: Storage Cleaner (`storage-cleaner`)

**Goal:** Port the `storage-cleaner` plugin from the old Steam Loader repo into the new Loadout repo, one plugin at a time. Port it *faithfully* (zero behavior regressions), simplify only where it is provably safe, and keep all of its code *inside the plugin* unless something is genuinely shared by 2+ already-migrated plugins.

## Locations
  - This plugin: `plugins/storage-cleaner/`
  - Port into: `plugins/storage-cleaner/`

---

**Layout** — `plugins/storage-cleaner/`:
- `package.json` — name `@loadout/plugin-storage-cleaner`, `"type": "module"`, deps, and a `plugin` field (the manifest now lives here, NOT in a separate `plugin.json`):
  ```json
  {
    "name": "@loadout/plugin-storage-cleaner",
    "version": "0.0.1",
    "type": "module",
    "plugin": {
      "id": "storage-cleaner",
      "name": "Storage Cleaner",
      "description": "…",
      "category": "…",
    }
  }
  ```

```bash
```

---

| Distro | Notes |
|---|---|

**Binary classification:**
  - `hyprctl` — Hyprland only.

---

## Porting procedure

1. **Copy** the source plugin tree `plugins/storage-cleaner/` from SOURCE into TARGET `plugins/storage-cleaner/`.
3. **Fold `plugin.json` into `package.json`.** Move the manifest fields into the `plugin` field of `package.json`, set `name` to `@loadout/plugin-storage-cleaner`, add `"type": "module"` and the real `dependencies`. Delete the standalone `plugin.json`.

---

|---|---|---|
| `plugin-storage` | ×8 | Inline: read/write JSON at `~/.config/loadout/plugins/storage-cleaner.json` (~30–50 LOC). |
| `sgdb-art` | ×2 | Inline. |

---

**Safe to cut / collapse:**

**NOT safe — do NOT touch:**

---

## Definition of Done

- [ ] `bun run typecheck`
- [ ] `bun run build`.
- [ ] **Plugin loads**: the loader logs `Loaded plugin: Storage Cleaner (storage-cleaner) …` (from `apps/loadout/src/loader/plugin-manager.ts`).

---

- **Plugin id:** `storage-cleaner`
- **Plugin name:** `Storage Cleaner`
- **Source path:** `linux-gaming-plugin-manager/plugins/storage-cleaner/`

srsholmes:
Closes #10.

## Summary

Faithful port of the `storage-cleaner` plugin from the steam-loader monorepo. Disk usage analyzer + Steam shader-cache / Proton compatdata cleanup, surfaced as an overlay plugin.

## Shared packages consumed

- `@loadout/ui` — `Button`, `Spinner`, `useBackend`, `mountComponent`, `PluginProvider`
- `@loadout/types` — `PluginBackend`, `EmitPayload`
- `@loadout/exec` — `run`, `runFull` (replaces `Bun.spawn`)
- `@loadout/steam-paths` — `getSteamAppsDir`

Plus `react`, `react-dom`, `react-icons/fa6` (`FaTrash`, `FaHardDrive`).

No new packages extracted. No removed-package deps to inline (the source plugin didn't use `plugin-storage`, `vdf`, etc — it stores no settings).

## Target

`overlay` only. Backend RPCs + a single React tree mounted into the overlay shell. No Steam-CEF injection, no homepage widget.

### RPC surface (unchanged)

- `getDiskUsage()` → `DiskPartition[]`
- `getShaderCacheSize()` → `{ total, totalFormatted, games[] }`
- `getCompatDataSize()` → `{ total, totalFormatted, games[] }`
- `getOrphanedData()` → `{ total, totalFormatted, entries[] }`
- `cleanShaderCache(appIds)` → `{ deleted, errors }` + emits `cacheCleared`
- `cleanCompatData(appIds)` → `{ deleted, errors }` + emits `cacheCleared`

## Inlined / extracted helpers

Pure logic was promoted from `backend.ts` and `app.tsx` into `lib/` per the contract:

- `lib/size.ts` — `formatSize`, `parseSizeToGB`, `bytesToGB`, `formatGB` (shared by backend + UI)
- `lib/appid.ts` — `isValidAppId` (path-traversal guard)
- `lib/parse-df.ts` — `parseDfOutput` + `DiskPartition` type
- `lib/parse-acf.ts` — `parseAcf` (Steam manifest key/value extract)

Each has a co-located `*.test.ts`. Backend now = I/O + RPC plumbing only.

## Filesystem paths touched

Declared in `plugin.permissions.filesystem`:

- `read:~/.local/share/Steam/steamapps` — `appmanifest_*.acf` scan, `du -sb` on shadercache/compatdata subdirs
- `write:~/.local/share/Steam/steamapps/shadercache` — `rm -rf <shadercache>/<appId>`
- `write:~/.local/share/Steam/steamapps/compatdata` — `rm -rf <compatdata>/<appId>`

## Commands declared

`["df", "du", "rm"]` — all universal coreutils, no bundling needed.

## Deviations from source

1. **Removed `plugin.json`** — manifest folded into `package.json` `plugin` field per Loadout contract.
2. **Replaced `createRoot` + `PluginProvider` boilerplate** in `app.tsx` with `mountComponent(...)` from `@loadout/ui` (both `mount` and `mountHeader`). Same observable behaviour.
3. **Extracted pure logic to `lib/`** as listed above. Removes inline duplicates of `parseSizeToGB`/`formatGB` from the UI and `formatSize`/`isValidAppId` from the backend.
4. **Tests converted to bun:test** — `app.spec.tsx` was vitest in the source; ported to `bun:test` with `mock.module("@loadout/ui", ...)` capturing the real module first (bun's `mock.module` isn't hoisted). Source's `backend.spec.ts` was already bun:test and renamed to `backend.test.ts`; spawn-stubbing added so tests don't rely on real filesystem state.

No RPC method names, signatures, or emitted-event payloads changed.

## Per-distro verdict

All three commands (`df`, `du`, `rm`) are universal coreutils, present on every targeted distro:

- SteamOS — likely
- CachyOS — likely
- Bazzite — likely
- Nobara — likely
- ChimeraOS — likely

Filesystem paths assume the canonical Steam install (`~/.local/share/Steam/...`). On Flatpak Steam the steam-paths package would need to grow Flatpak detection — same gap every other steam-paths consumer has; not a regression.

## CI gates

- `bun run typecheck` clean
- `bun run lint` 35 warnings / 0 errors (baseline 35 — delta 0)
- `bun run check:specs` OK
- `bun run test:backend` 943 pass / 0 fail
- `bun run test:ui` 168 pass / 0 fail
- `bun run build` OK


srsholmes:
## PR #71 — storage-cleaner migration review

### Code quality
- **Pure-logic extraction is well done.** `formatSize`, `parseSizeToGB`, `bytesToGB`, `formatGB`, `parseDfOutput`, `parseAcf`, `isValidAppId` all moved out of inline `backend.ts` into `lib/{size,parse-df,parse-acf,appid}.ts` with co-located tests. Faithful, simplify-without-regression.
- **Vestigial `parseAppManifests` lives in `backend.ts:26`.** It's pure-ish (I/O is `readdir` + `Bun.file().text()`, no `this`), and it's a 17-LOC helper called from three RPCs. Defensible to leave inline (single plugin, tight coupling), but the migration prompt's "pure stuff = lib/" rule could push it to `lib/manifests.ts` with an injectable reader. Not a blocker — flag only.
- **`getDirSize` (backend.ts:49)** — same shape: 8-LOC `du -sb` wrapper, used only by this plugin. Inlined is fine, but if `getDirSize` ever grows it's the natural lib candidate.
- **`backend.test.ts:55`** uses `spyOn(Bun, "spawn")` — preferred per `docs/test-mock-contamination.md`. Good choice.
- **`app.spec.tsx:11–24`** uses `mock.module("@loadout/ui", ...)` with `import * as actualUi` captured first — correct pattern per the migration prompt for JSX-imported SDK surface.
- **No leftover `@steam-loader/*` imports**, no `sudo`/`pkexec` wrappers, no `console.log` debris beyond the lifecycle markers (which mirror the source).
- **DRY:** zero duplication against shared packages. Plugin doesn't persist state, so no `@loadout/plugin-storage` need. Uses `mountComponent` + `useBackend` from `@loadout/ui` correctly (no inline `createRoot + PluginProvider` boilerplate).

### Bundled-binary quality
- No `bundled_bins` — uses universal `df`, `du`, `rm`. Skip.

### Lint-warnings count
Baseline (main): **35** — PR: **35** — Delta: **0**. Zero new errors. Zero `as any` / `: any` casts added in this plugin's files.

- `bun run check:specs`: **pass** ("every plugin entry point ≥ 100 LOC has a sibling spec").
- `bun run test:backend`: **943 pass, 0 fail**. `bun run test:ui`: **214 pass, 0 fail**. Plugin's own 46 backend + 8 UI specs all green.
- Pure-logic coverage: `lib/size.test.ts` (16 cases incl. boundary 100GB / 10GB), `lib/parse-df.test.ts` (5 cases incl. dedup + malformed), `lib/parse-acf.test.ts` (6 cases incl. null returns), `lib/appid.test.ts` (path traversal + command injection cases). All extracted helpers covered.
- Backend RPC coverage: lifecycle, appId validation (path traversal + injection), `emit` contract on both clean methods, `getDiskUsage` happy + empty, sort-order assertions for `getShaderCacheSize` / `getCompatDataSize` / `getOrphanedData`.
- Gap: `parseAppManifests` (backend.ts:26) is exercised only transitively. If it moves to `lib/`, dedicated coverage for the silent-skip branches becomes possible without spy gymnastics.

|---|---|---|
| `df` | ✅ Universal | coreutils on every target |
| `du` | ✅ Universal | coreutils on every target |
| `rm` | ✅ Universal | coreutils on every target |
| `read:~/.local/share/Steam/steamapps` | ⚠️ Steam-conditional | Default Steam path. Flatpak Steam uses `~/.var/app/com.valvesoftware.Steam/.local/share/Steam/...` — `getSteamAppsDir()` from `@loadout/steam-paths` is what actually resolves; the permissions string is documentation only |
| `write:.../shadercache` | ⚠️ Steam-conditional | Same |
| `write:.../compatdata` | ⚠️ Steam-conditional | Same |

**Verdict (out of the box):**
- SteamOS (stock Deck): ✅ likely — `df`/`du`/`rm` present; default Steam install path matches.
- CachyOS:              ✅ likely — coreutils + Steam package present.
- Bazzite:              ✅ likely — atomic but coreutils universal; Steam at default path on KDE Plasma image.
- Nobara:               ✅ likely — gaming-tooling distro, Steam preinstalled.
- ChimeraOS:            ✅ likely — handheld console image, Steam default.

Plugin degrades cleanly when paths are missing (every `readdir` is wrapped in `try { ... } catch {}`).

- **`getDiskUsage` only reports `/` and `$HOME`** (`backend.ts:74`), then dedupes by filesystem. On systems where `/home` is a separate disk AND a third game-library mount (e.g. `/run/media/mmcblk0p1` on Deck SD card, or any extra steam-library drive) holds shader/compat data, that drive never appears in the disk summary — but its shader/compat sizes still feed the stacked bar via `getSteamAppsDir()`. Result: "used" GB on the bar can exceed the displayed `total` GB on multi-library setups. Source plugin has the same bug; not a migration regression.
- **`parseDfOutput` (lib/parse-df.ts:22)** splits on `/\s+/` and indexes `parts[0..5]`. `df -h` on filesystems with a space in the mount label (rare but possible — e.g. user-mounted volumes) shifts columns. No bug in practice on Steam paths; flag only.
- **`parseAcf` (lib/parse-acf.ts:14)** matches the *first* `"appid"` and `"name"` in the file. ACF `UserConfig`/`MountedDepots` sub-blocks don't have `"name"` keys, so collisions are unlikely, but the parser isn't structurally aware. Carried over from source.
- **`cleanShaderCache` / `cleanCompatData` shell out to `rm -rf`** instead of `rm.rm`/`fs.rm` (Bun has `fs.rm(..., { recursive, force })`). The appId regex `/^\d+$/` (lib/appid.ts:6) makes this safe today; flagging because `rm -rf` over a subprocess boundary is a sharper edge than the `fs` API for the same operation.
- **`getOrphanedData` (backend.ts:149)** re-walks `shadercache` and `compatdata` with `du -sb` on every call. On a Deck with 200+ orphaned entries each `du -sb` is a separate fork. Not a regression — source plugin does the same — but worth flagging as a future caching candidate.

srsholmes:
## Review fixes — all 5 "Potential bugs" addressed

Pushed \`b14809f\`. Summary per item:

1. **\`getDiskUsage\` multi-library** — Now queries \`@loadout/steam-paths#getLibraryPaths()\` and feeds every library path + \`/\` + \`\$HOME\` to \`df -h\`, deduped by filesystem. Missing libraries (e.g. unmounted SD card) are filtered via \`fs.stat\` first, so \`df\` doesn't spam stderr.
2. **\`parseDfOutput\` multi-word mountpoints** — \`parts.slice(5).join(" ")\` instead of \`parts[5]\`. Test added: \`/run/media/My SD Card\`.
3. **\`parseAcf\` brace-aware** — Rewrote as a tokenizer + depth tracker. Only top-level keys of \`AppState\` are read; nested blocks (\`InstalledDepots\`, \`UserConfig\`) can no longer shadow the real \`appid\` / \`name\`. Test added with a nested-\`appid\` collision.
4. **\`rm -rf\` → \`fsp.rm\`** — Dropped the \`runFull(["rm", "-rf", target])\` subprocess. \`{ recursive: true, force: true }\` covers the missing-dir case. Test now spies \`fsp.rm\` directly per \`docs/test-mock-contamination.md\` — also means the test can never delete real Steam files on a dev machine. Cleanup loops collapsed into a private \`_cleanCache(type)\` helper.
5. **\`du -sb\` batching** — New pure \`lib/parse-du.ts\` for the multi-line output. \`getShaderCacheSize\` / \`getCompatDataSize\` / \`getOrphanedData\` now run a single \`du -sb path1 path2 ...\` per RPC; on a Deck with 171 orphaned entries (verified locally) that's 1 fork vs 171.

### Tests
- 69 pass, 0 fail (54 PR + 5 manifests + 4 parse-du + new parseAcf/parseDf coverage).
- \`bun run typecheck\` clean.
- \`bun run lint\` clean.

### Cross-distro verification

**Steam Deck (SteamOS 3.7)** — end-to-end:
- Plugin loads cleanly (\`[storage-cleaner] Plugin loaded\`, \`[backend=yes, frontend=yes]\`).
- \`getLibraryPaths\` finds both default + extra library on this Deck.
- \`getDiskUsage\` correctly returns 2 distinct filesystems (root + home).
- \`getShaderCacheSize\` finds 118 games, 22.56 GB.
- \`getOrphanedData\` finds 171 orphans in a single batched \`du\`.

**Bazzite (and other glibc/coreutils 9.x distros)** — analysis:
- \`df -h\`, \`du -sb\` are GNU coreutils, identical behavior. \`-b\` (apparent bytes) is portable across both.
- \`fsp.rm\`: Node API, identical.
- \`getLibraryPaths\`: reads \`~/.local/share/Steam/steamapps/libraryfolders.vdf\` — same on Bazzite native Steam.
- Locale: \`df -h\` size strings (\`1.5T\` vs \`1,5T\`) are stored opaque and displayed pass-through, so non-en_US LC_NUMERIC is tolerated.
- **Known carry-over limitation**: \`@loadout/steam-paths\` hardcodes the native Steam dir; Flatpak Steam (rare on Bazzite gaming images, common on stock Fedora) returns empty results. Pre-existing across all loadout plugins; out of scope for this PR.

Apex (Bazzite-derived) testing welcome — should be a straight pull-build-install. The only Bazzite-specific install difference is the binary lands at \`/usr/local/bin/loadout\` per \`docs/install-locations.md\` and SELinux gets relabeled, both handled by \`install-local.sh\` already.

PR #73, opened 2026-05-28
Migrate Disable Controller Input plugin (disable-controller-input)

### LINKED ISSUE #13: Migrate plugin: Disable Controller Input (disable-controller-input)
> **Auto-detected** — source: `linux-gaming-plugin-manager/plugins/disable-controller-input/` · UI: overlay (app.tsx) · removed-package deps to inline: plugin-storage · subprocess: .

# Migrate one plugin: Disable Controller Input (`disable-controller-input`)

**Goal:** Port the `disable-controller-input` plugin from the old Steam Loader repo into the new Loadout repo, one plugin at a time. Port it *faithfully* (zero behavior regressions), simplify only where it is provably safe, and keep all of its code *inside the plugin* unless something is genuinely shared by 2+ already-migrated plugins.

## Locations
  - This plugin: `plugins/disable-controller-input/`
  - Port into: `plugins/disable-controller-input/`

---

**Layout** — `plugins/disable-controller-input/`:
- `package.json` — name `@loadout/plugin-disable-controller-input`, `"type": "module"`, deps, and a `plugin` field (the manifest now lives here, NOT in a separate `plugin.json`):
  ```json
  {
    "name": "@loadout/plugin-disable-controller-input",
    "version": "0.0.1",
    "type": "module",
    "plugin": {
      "id": "disable-controller-input",
      "name": "Disable Controller Input",
      "description": "…",
      "category": "…",
    }
  }
  ```

```bash
```

---

| Distro | Notes |
|---|---|

**Binary classification:**
  - `hyprctl` — Hyprland only.

---

## Porting procedure

1. **Copy** the source plugin tree `plugins/disable-controller-input/` from SOURCE into TARGET `plugins/disable-controller-input/`.
3. **Fold `plugin.json` into `package.json`.** Move the manifest fields into the `plugin` field of `package.json`, set `name` to `@loadout/plugin-disable-controller-input`, add `"type": "module"` and the real `dependencies`. Delete the standalone `plugin.json`.

---

|---|---|---|
| `plugin-storage` | ×8 | Inline: read/write JSON at `~/.config/loadout/plugins/disable-controller-input.json` (~30–50 LOC). |
| `sgdb-art` | ×2 | Inline. |

---

**Safe to cut / collapse:**

**NOT safe — do NOT touch:**

---

## Definition of Done

- [ ] `bun run typecheck`
- [ ] `bun run build`.
- [ ] **Plugin loads**: the loader logs `Loaded plugin: Disable Controller Input (disable-controller-input) …` (from `apps/loadout/src/loader/plugin-manager.ts`).

---

- **Plugin id:** `disable-controller-input`
- **Plugin name:** `Disable Controller Input`
- **Source path:** `linux-gaming-plugin-manager/plugins/disable-controller-input/`

srsholmes:
Closes #13.

Ports the `disable-controller-input` plugin from the old steam-loader repo into Loadout. Faithful port — same RPC method names + signatures, same `controllersChanged` event payload, same persisted-state shape on disk, same 2s reconcile cadence.

## Shared packages consumed

- `@loadout/types` (`PluginBackend`, `EmitPayload`)
- `@loadout/ui` (`Alert`, `Button`, `IconButton`, `Spinner`, `Toggle`, `mountComponent`, `useBackend`)
- `@loadout/exec` (`runFull` with 5s `timeoutMs` ceiling — Audit D-009)
- `@loadout/plugin-storage` (`readPluginStorage` / `writePluginStorage` for the per-device intent cache)

No new shared package introduced.

## Target

Overlay only (`app.tsx`). Backend drives InputPlumber over DBus; nothing touches Steam's CEF UI, so no `@loadout/steam-cdp` dependency.

## Inlined helpers

Pure helpers promoted to `lib/parse.ts` with sibling tests:
- `djb2(name)` — 32-bit hash used as a stable on-disk device id across reconnects.
- `parseStringProp` — parses `s "value"` busctl output (with `\"` / `\\` unescaping).
- `parseObjectPathArrayProp` — parses `ao N "/p1" "/p2"` busctl output.
- `pickCompositePaths` — filters `busctl tree --list` output down to top-level `CompositeDevice<N>` paths.

Mirrors the contract's "pure stuff lives in `lib/`, backend is I/O + RPC plumbing".

## Subprocesses + permissions

| Permission | Value |
|---|---|
| `permissions.commands` | `["busctl"]` |
| `permissions.filesystem` | (none — no `fs` access; everything is DBus-mediated) |
| `permissions.network` | (none) |

Every external call goes through `runFull(["busctl", "--system", "--no-pager", …])`. No `sudo`/`pkexec` wrappers — backend runs as root.

## Deviations from source

1. **Header-as-card lockout warning → `Alert`.** The source rendered the "you'll need a keyboard/mouse/touchscreen to re-enable" warning as an inline flex row with a hand-rolled `--color-warning` triangle icon. Loadout has the `Alert` component for exactly this; swapped to `Alert variant="warning" title="Lock-out warning"`. Same copy, same intent, same colour token — cleaner shape and consistent with how `display-settings` / `fan-control` render banners.
2. **`mountComponent` instead of inline `createRoot + PluginProvider`.** Per the contract's "no inlined `createRoot + PluginProvider` boilerplate" — matches the display-settings reference.
3. **Private methods prefixed with `_`** (`_reconcile`, `_persist`) so they don't become RPC endpoints. The source named them `reconcile` / `persist` without an underscore; non-load-bearing rename to match the loader's wire-surface convention.
4. **Tests ported vitest → bun:test.** App spec uses `mock.module("@loadout/ui", …)` capturing `actualUi` before the mock (bun's `mock.module` isn't hoisted). Backend tests use `spyOn(Bun, "spawn")` and `spyOn(pluginStorage, "readPluginStorage" | "writePluginStorage")` instead of the source's `mock.module("@steam-loader/plugin-storage", …)` — `spyOn` is the recommended path for shared modules per `docs/test-mock-contamination.md`.

No behaviour change — same RPC surface (`listControllers`, `refreshControllers`, `setDisabled`, `forgetController`), same emitted event (`controllersChanged`), same `State` schema on disk, same `BUSCTL_TIMEOUT_MS = 5000` audit fix.

## Cross-distro check

| Distro | Verdict | Notes |
|---|---|---|
| SteamOS (stock Deck) | ✅ likely | InputPlumber ships with SteamOS Holo; `busctl` is universal. |
| CachyOS | ⚠️ partial | InputPlumber is in the AUR but not installed by default. Plugin detects-and-degrades: `inputPlumberAvailable()` returns false → UI shows the "InputPlumber not detected" banner with `systemctl status inputplumber.service` hint. |
| Bazzite | ✅ likely | InputPlumber is preinstalled on Bazzite + Bazzite-Deck. |
| Nobara | ⚠️ partial | InputPlumber not in the base image; same detect-and-degrade banner shown. |
| ChimeraOS | ✅ likely | InputPlumber is part of the handheld console image. |

Plugin gracefully no-ops when InputPlumber is absent — cache still hydrates from disk, RPC returns `unavailable: true`, UI shows a banner with re-check button. No crash path.

## LOC delta vs source

| File | Source | Loadout | Δ |
|---|---|---|---|
| `backend.ts` | 503 | 459 | −44 (pure helpers moved to `lib/parse.ts`) |
| `app.tsx` | 303 | 280 | −23 (`mountComponent` collapses two `createRoot+PluginProvider` blocks) |
| `backend.spec.ts` → `backend.test.ts` | 479 | 504 | +25 (spyOn-based storage mock vs `mock.module`) |
| `app.spec.tsx` | 110 | 126 | +16 (bun:test + `actualUi` capture preamble) |
| `lib/parse.ts` (new) | — | 62 | +62 |
| `lib/parse.test.ts` (new) | — | 113 | +113 |
| `package.json` (manifest folded in) | 25 | 23 | −2 |
| `plugin.json` (deleted) | 9 | 0 | −9 |
| **Total** | **1429** | **1544** | **+115** (mostly the new `lib/parse.test.ts`) |

## CI gates

- [x] `bun run typecheck` — clean
- [x] `bun run lint` — 0 errors, 35 warnings (matches baseline; this plugin adds 0)
- [x] `bun run check:specs` — `OK — every plugin entry point ≥ 100 LOC has a sibling spec`
- [x] `bun run test:backend` — 940 pass, 0 fail (includes this plugin's 11 backend + 18 lib tests)
- [x] `bun run test:ui` — 165 pass, 0 fail (includes this plugin's 5 UI tests)
- [x] `bun run build` — clean


srsholmes:
## PR #73 — disable-controller-input migration review

### Code quality
- **Reconcile cost grows linearly with disabled devices**: `backend.ts:413-435` issues `getTargetPaths` + one `getTargetKind` per target every 2 s for every cached `disabled: true` device, even when nothing changed. With N silenced pads and M targets each, that's N + N*M busctl calls per tick. Reasonable now, scales poorly.
- **`busCtl` re-walks fresh kinds even when nothing has changed**: `_reconcile` snapshots `liveKinds` and re-issues `SetTargetDevices` on every tick where the daemon happens to expose targets. There's no "is this already silenced?" early exit beyond the `currentTargets.length > 0` gate. Fine for correctness; noisy on the bus.
- **Pure helpers cleanly extracted**: `lib/parse.ts` (djb2, parseStringProp, parseObjectPathArrayProp, pickCompositePaths) is exactly the right granularity. No further extraction needed; backend.ts (459 LOC) is mostly RPC + reconcile orchestration.
- **Conventions met**: `mountComponent` for both views (`app.tsx:279-280`), `useBackend` from `@loadout/ui`, `runFull` from `@loadout/exec` (no raw `Bun.spawn`), `@loadout/plugin-storage` for persistence, no `sudo`, no `@steam-loader/*` remnants. Imports clean.
- **No cross-plugin imports.** Plugin-seal rule satisfied.

### Bundled-binary quality
No `bundled_bins` declared. Plugin depends on `busctl` (universal on systemd distros) and the InputPlumber DBus service (not bundled — flagged below for cross-distro impact). Appropriate decision — InputPlumber is a system daemon, not bundle-able as a single binary.

### Lint-warnings count
Baseline (main): 35 — PR: 35 — Delta: 0
- 13 `as any/unknown` casts in this plugin, all confined to `backend.test.ts` for `spyOn` return shapes and one `as any` in `app.spec.tsx` for the mock module shim. Each cast has an inline `eslint-disable-next-line` comment with a rationale. No new lint warnings introduced (the casts pre-existed in the test setup patterns).

- `bun run check:specs`: **pass** (every entry point ≥ 100 LOC has a sibling spec).
- `bun run test:backend`: **940 pass, 0 fail** across 61 files.
- `backend.test.ts` covers: unavailable detection, cache hydration, offline `setDisabled`, online `setDisabled` (savedKinds snapshot + null silencing), re-enable restore, `forgetController`, **and the D-009 busctl-timeout regression** (10s timeout fixture; verifies kill fires within 7s).
- `lib/parse.test.ts` covers all four pure functions with edge cases (empty input, escaped quotes, malformed signatures, multi-path arrays).
- Gap: no test for the `_reconcile` re-silence path (lines 412-435) — the "InputPlumber restarted and restored defaults, we re-snapshot and re-null" loop. Worth a fixture before this lands.

|---|---|---|
| `busctl` | ✅ Universal | systemd default on all five targets |
| `org.shadowblip.InputPlumber` DBus service | ⚠️ Session-conditional | Preinstalled on SteamOS, Bazzite, ChimeraOS; NOT on stock CachyOS/Nobara — user installs `inputplumber-git` from AUR or `inputplumber` from Fedora COPR |

**Verdict (out of the box):**
- SteamOS (stock Deck): ✅ likely — InputPlumber ships in the Deck image
- CachyOS: ⚠️ partial — InputPlumber not in base; plugin renders "InputPlumber not detected" banner with install hint (no crash, just inert)
- Bazzite: ✅ likely — InputPlumber preinstalled on Bazzite-Deck; on desktop Bazzite usually present via gaming meta-package
- Nobara: ⚠️ partial — InputPlumber not in stock image; user installs via COPR; banner correctly surfaces
- ChimeraOS: ✅ likely — InputPlumber default in handheld image

The "InputPlumber not detected" Alert at `app.tsx:100-131` handles the missing-daemon case gracefully — banner + Re-check button, no error throwing. Reconcile loop also self-heals when the daemon comes up later (`backend.ts:364-369`).

- **`hardRefresh` callback dependency mismatch**: `app.tsx:40-43` declares `hardRefresh` with `[call]` dependency but uses it inside Refresh button click handlers. Not a bug per se, but stale-callback risk if `call` identity ever changes mid-render (it doesn't currently). Cosmetic.
- **Default kinds may not match the device's actual capabilities** (`backend.ts:45`): `DEFAULT_KINDS = ["xb360", "mouse", "keyboard"]` is hardcoded. If the user disables a controller while unplugged (no live kinds to snapshot), then later re-enables, they get the default set — which may not match what InputPlumber would have spawned (e.g. `deck` mode for Steam Deck built-in, or `ds5` for a DualSense). Cosmetic until reported.
- **`_reconcile` re-silence races user toggle** (`backend.ts:413-435` vs `backend.ts:294-306`): if a user toggles `disabled = true` and a reconcile tick fires before `_persist` completes, the tick sees the device with the live target list and **re-snapshots** the just-changed targets into `savedKinds`. Probably benign (same kinds), but not serialized. Worth a `reconciling` mutex check inside RPC methods or an `await this._persist()` in `setDisabled` before the next tick reads state. The `reconciling` guard at line 359 only prevents overlapping reconciles, not RPC-vs-reconcile.
- **`forgetController` re-enable best-effort, no error surfaced** (`backend.ts:338-341`): if `setTargetKinds` fails inside `forgetController`, the return value is discarded and the device is still removed from the cache. The silenced target stays orphaned on the bus until next InputPlumber restart. The UI sees a successful forget. Worth at least logging the failure.
- **`pickCompositePaths` won't match a child path mid-line**: `lib/parse.ts:51-52` regex is `^...CompositeDevice\d+$` (no trailing chars). Confirmed via test, but worth noting that the `busctl tree --list` format is undocumented and could change. Defensive enough.

PR #75, opened 2026-05-29
Migrate LSFG-VK plugin (lsfg-vk)

### LINKED ISSUE #23: Migrate plugin: LSFG-VK (lsfg-vk)
> **Auto-detected** — source: `linux-gaming-plugin-manager/plugins/lsfg-vk/` · UI: overlay (app.tsx) · removed-package deps to inline: vdf · subprocess: .

# Migrate one plugin: LSFG-VK (`lsfg-vk`)

**Goal:** Port the `lsfg-vk` plugin from the old Steam Loader repo into the new Loadout repo, one plugin at a time. Port it *faithfully* (zero behavior regressions), simplify only where it is provably safe, and keep all of its code *inside the plugin* unless something is genuinely shared by 2+ already-migrated plugins.

## Locations
  - This plugin: `plugins/lsfg-vk/`
  - Port into: `plugins/lsfg-vk/`

---

**Layout** — `plugins/lsfg-vk/`:
- `package.json` — name `@loadout/plugin-lsfg-vk`, `"type": "module"`, deps, and a `plugin` field (the manifest now lives here, NOT in a separate `plugin.json`):
  ```json
  {
    "name": "@loadout/plugin-lsfg-vk",
    "version": "0.0.1",
    "type": "module",
    "plugin": {
      "id": "lsfg-vk",
      "name": "LSFG-VK",
      "description": "…",
      "category": "…",
    }
  }
  ```

```bash
```

---

| Distro | Notes |
|---|---|

**Binary classification:**
  - `hyprctl` — Hyprland only.

---

## Porting procedure

1. **Copy** the source plugin tree `plugins/lsfg-vk/` from SOURCE into TARGET `plugins/lsfg-vk/`.
3. **Fold `plugin.json` into `package.json`.** Move the manifest fields into the `plugin` field of `package.json`, set `name` to `@loadout/plugin-lsfg-vk`, add `"type": "module"` and the real `dependencies`. Delete the standalone `plugin.json`.

---

|---|---|---|
| `plugin-storage` | ×8 | Inline: read/write JSON at `~/.config/loadout/plugins/lsfg-vk.json` (~30–50 LOC). |
| `sgdb-art` | ×2 | Inline. |

---

**Safe to cut / collapse:**

**NOT safe — do NOT touch:**

---

## Definition of Done

- [ ] `bun run typecheck`
- [ ] `bun run build`.
- [ ] **Plugin loads**: the loader logs `Loaded plugin: LSFG-VK (lsfg-vk) …` (from `apps/loadout/src/loader/plugin-manager.ts`).

---

- **Plugin id:** `lsfg-vk`
- **Plugin name:** `LSFG-VK`
- **Source path:** `linux-gaming-plugin-manager/plugins/lsfg-vk/`

srsholmes:
Closes #23

## Summary
Faithful port of the LSFG-VK plugin (Lossless Scaling for Linux via Vulkan layer) from steam-loader to Loadout. Zero behaviour regressions: same RPC method names, event names, payload shapes, and on-disk protocol with the PancakeTAS/lsfg-vk Vulkan layer (`~/lsfg` wrapper + `~/.config/lsfg-vk/conf.toml`).

## Shared packages consumed
- `@loadout/ui` — `mountComponent`, `mountHeaderStub`, `PluginProvider`, `Spinner`, `Button`, `Toggle`, `Slider`, `Select`, `SegmentedItem`, `TextInput`, `SearchField`, `IconButton`, `HeaderBackButton`, `PluginHeader`, `GameCard`, `fuzzySearchGames`, `useBackend`, `useCurrentGame`, `useFocusable`.
- `@loadout/types` — `PluginBackend`, `EmitPayload`, `PluginLogger`, and `GameInfo` / `GameCollection` (the `__core:game-library` RPC contract).
- `@loadout/exec` — `run`, `runFull`, `commandExists`, `spawn` (gated on the declared `permissions.commands` list).
- `@loadout/plugin-storage` — `readPluginStorage` / `writePluginStorage` for settings persistence; replaces the source's hand-rolled `~/.config/steam-loader/lsfg-vk/config.json`.
- `@loadout/vdf` — `hasLaunchToken` (consumed by the apply-to-game picker to decide Apply vs. Remove).

## Target
- `target.type: "overlay"`. Main settings UI in `app.tsx`; a `mountHomeWidget` exposes the QAM 4-cell multiplier picker; `mountHeader = mountHeaderStub` reserves the topbar slot and the dynamic header content is portaled via `<PluginHeader>` from inside `mount()`.

## `__core:game-library` rewire
The apply-to-game `GamePicker.tsx` previously called `useBackend("game-browser")` (`getGames`, `getCollections`). That plugin was retired; rewired to `useBackend("__core:game-library")` and replaced the local `GameInfo` / `CollectionEntry` types with `GameInfo` / `GameCollection` from `@loadout/types`. Matches the pattern PR #65 / #66 / launch-options refactor used.

## Inlined helpers
Pure TOML + wrapper-script renderers hoisted from `backend.ts` into `lib/render-config.ts` with a sibling `render-config.test.ts` (per the "pure logic in `lib/`" rule). UI hooks `use-flash-status`, `use-lsfg-manager`, `use-picker-filters` ported verbatim — all under the 100-LOC spec floor, so no required sibling tests beyond the `app.spec.tsx` integration coverage.

## Subprocesses + filesystem (per-distro classification)
**`permissions.commands`** (declared, gated at `@loadout/exec`):
- `unzip` — universal ✅ on SteamOS / CachyOS / Bazzite / Nobara / ChimeraOS.
- `wl-copy` (wl-clipboard) — ⚠️ session-conditional; backend tries it first, falls back to `xclip`, returns a structured error if neither is present.
- `xclip` — ⚠️ session-conditional; same fallback chain.
- `vulkaninfo`, `vkcube` — ⚠️ vulkan-tools; backend detects via `commandExists` and surfaces an install hint when missing. Stock SteamOS / Bazzite-Deck have these; Nobara/CachyOS users may need to install vulkan-tools.

**`permissions.filesystem`**:
- `~/.local/lib`, `~/.local/share/vulkan/implicit_layer.d`, `~/.config/lsfg-vk`, `~/lsfg` — Vulkan layer install targets. Universal.
- `~/.local/share/Steam/steamapps/common/Lossless Scaling` — Lossless.dll auto-detection path. Read-only probe; if the user installed Lossless Scaling on a different Steam library they set a custom path via the UI.

**`permissions.network`**: `api.github.com`, `github.com`, `objects.githubusercontent.com` — GitHub Releases install fetch.

## Per-distro verdict
| Distro | Verdict | Notes |
|---|---|---|
| SteamOS (stock Deck) | ✅ likely | vulkan-tools available in steam-runtime; gamescope provides Wayland → `wl-copy` fallback works. |
| CachyOS | ✅ likely | All commands available via repos / AUR. |
| Bazzite | ✅ likely | `rpm-ostree install vulkan-tools` documented in the diagnostics card hint. |
| Nobara | ✅ likely | Same as Bazzite re: vulkan-tools, but mutable. |
| ChimeraOS | ✅ likely | Vulkan + handheld tooling preinstalled. |

The plugin degrades gracefully on every distro: missing `vulkaninfo` / `vkcube` → diagnostic card returns a structured error with the install command; missing clipboard tool → Copy button surfaces "install wl-clipboard or xclip" message.

## Deviations from source
- Persistence path moved from `~/.config/steam-loader/lsfg-vk/config.json` to `~/.config/loadout/plugins/lsfg-vk.json` (via `@loadout/plugin-storage`). Same JSON shape; users re-installing the plugin start fresh on first launch.
- `GamePicker` consumes `__core:game-library` instead of the now-retired `game-browser` plugin RPC.
- `GameInfo` / `GameCollection` come from `@loadout/types`, not a local copy.
- Mount boilerplate (`createRoot` + `PluginProvider`) replaced with `mountComponent` / `mountHeaderStub` factories from `@loadout/ui` — same behaviour, less duplication. The QAM widget keeps its one-off `mount()` because it must NOT reserve a topbar slot.
- Backend tests converted to bun:test; fetch error-path mocked with `spyOn(globalThis, "fetch")` instead of `globalThis.fetch = …` (per the testing guidance to prefer spies).

## Test plan
- [x] `bun run typecheck` clean.
- [x] `bun run lint` — 35 warnings, 0 errors (baseline matches main exactly; no new warnings introduced).
- [x] `bun run check:specs` passes.
- [x] `bun run test` — 167 UI + 942 backend tests pass; the 1 pre-existing `playtime` backend failure on main is unrelated.
- [x] `bun run build` produces the Electrobun overlay bundle.
- [x] Plugin tree validated against `scripts/check-plugin-specs.sh` (MIN_LOC = 100): `backend.ts` (530 LOC) → `backend.test.ts` ✓; `app.tsx` (199 LOC) → `app.spec.tsx` ✓; `lib/render-config.ts` (66 LOC) → covered by `render-config.test.ts` for the pure logic rule even though it's under the floor.
- [ ] End-to-end: install from the overlay, apply to a Steam game, verify wrapper + TOML on disk (manual; needs the overlay running in gamescope).


srsholmes:
## PR #75 — LSFG-VK migration review

### Code quality
- **Solid decomposition.** 530-LOC `backend.ts` cleanly extracts pure logic (`renderTomlConfig`, `renderWrapperScript`, `tomlString`) into `lib/render-config.ts` with co-located unit tests. UI is decomposed into 9 stateless cards + 3 hooks (`useFlashStatus`, `useLsfgManager`, `usePickerFilters`). `app.tsx` is 199 LOC of pure composition, no inline RPC plumbing.
- **Conventions clean.** All imports stay on `@loadout/{ui,types,exec,plugin-storage,vdf}` + react/react-icons. No `@steam-loader/*`, no `Bun.spawn` outside `@loadout/exec` (the one `spawn` import on `backend.ts:2` is from `@loadout/exec` — fine). No `sudo` / `pkexec`. No legacy / migration code. Storage is `readPluginStorage` / `writePluginStorage`. Mount uses `mountComponent` / `mountHeaderStub`.
- **Test style.** Uses `spyOn(globalThis, "fetch")` for the install error path (preferred per spec); UI spec uses `mock.module` for the backend handle, which is fine under `--isolate`.
- **Minor**: `GamePicker.tsx:127` keeps a `collections` state that's only `void`-ed on line 175 (`void collections;` to silence unused-var). The setter is still called for the `onCollectionsLoaded` callback. Not a bug, but the state slot itself is dead — the deduped collections list is already lifted to `usePickerFilters` and the picker doesn't render it. Could drop the local `useState<GameCollection[]>` entirely.
- **Minor**: `backend.ts:7` imports `tmpdir` from `node:os` — used twice for download/extract paths. Fine, just noting it's the only `node:os` consumer besides `homedir`.

### Bundled-binary quality
- **No `bundled_bins`** — by design. The Vulkan layer (`liblsfg-vk.so` + JSON) is fetched at runtime from `PancakeTAS/lsfg-vk` GitHub releases (GPL-3.0, can't bundle BSD-3 plugin), and Lossless.dll is a paid Steam app (user-supplied). `NOTICE` correctly documents the GPL-3.0 runtime dep is downloaded, not bundled. Appropriate.

### Lint-warnings count
Baseline (main): **35** — PR: **35** — Delta: **0**, errors: **0**, new `as any` casts in this PR: **0**. Clean.

- `check:specs`: **pass**.
- `backend.test.ts` (294 LOC): covers defaults, `updateSettings` merge + persist + emit, install/getStatus reports, DLL detection (default + custom + clear), TOML/wrapper integration (real file writes), and the install error path via fetch spy. Strong.
- `lib/render-config.test.ts` (99 LOC): covers `tomlString` escaping, default/partial settings, multiplier=0 clamp, special-char DLL paths, wrapper script shebang + LSFG_PROCESS + verbose exports. Strong.
- **Pure helper candidates left in `backend.ts` (could move to `lib/`, but none are forced moves):**
  - `_findExtractedFiles` (backend.ts:459) — iterates a directory looking for two filenames. Pure-ish (uses `readdir`), would test fine with a tmpdir fixture, but it's I/O-bound so the case for extraction is weak.
  - `_resolveLatestNoUiAsset` (backend.ts:378) — the asset selector (`release.assets.find(a => a.name === ASSET_NAME)`) is a one-liner against a network fetch. Not worth extracting alone.
- **UI tests** (`app.spec.tsx`, 203 LOC): covers the loading / not-installed / installed paths with `mock.module` for `@loadout/ui`'s `useBackend`. Reasonable.

|---|---|---|
| `unzip` | ⚠️ | Universal on SteamOS / ChimeraOS / CachyOS / Nobara; **missing by default on Bazzite** (atomic; needs `rpm-ostree install unzip` or a layered package). |
| `wl-copy` | ⚠️ | Wayland clipboard tool. Preinstalled on Bazzite/Nobara KDE Wayland and most CachyOS sessions; **not preinstalled on SteamOS Gaming Mode** (gamescope is its own compositor — wl-clipboard isn't part of the stock image). Backend has a fallback to `xclip`, so behavior degrades to "Copy fails with a helpful message" rather than crashing — good. |
| `xclip` | ⚠️ | X11 clipboard. Same story — not on stock SteamOS / Bazzite-Deck. Fallback chain is correct. |
| `vulkaninfo` / `vkcube` | ⚠️ | From `vulkan-tools`. Not preinstalled on stock SteamOS or Bazzite. Backend detects absence via `commandExists` and returns a helpful "install vulkan-tools" message instead of throwing — correct. |
| `~/.local/lib`, `~/.local/share/vulkan/implicit_layer.d`, `~/.config/lsfg-vk`, `~/lsfg` | ✅ | User-home paths, universally writable. |
| `~/.local/share/Steam/steamapps/common/Lossless Scaling` | ⚠️ | Default Steam library only. Backend handles non-default libraries via the custom DLL path field — good. |
| `api.github.com`, `github.com`, `objects.githubusercontent.com` | ✅ | Universal. |

**Verdict (out of the box):**
- **SteamOS (stock Deck):** ⚠️ partial — install path works (`unzip` is on SteamOS); clipboard Copy + vkcube diagnostics fail with a "install X" message until the user layers `vulkan-tools` and `wl-clipboard` (or runs in Desktop Mode which has them). Core flow (install layer + apply to game) works.
- **CachyOS:** ✅ likely — all tools available or one `pacman -S` away; KDE/Wayland clipboard works out of the box.
- **Bazzite:** ⚠️ partial — `unzip` is **not in the stock Bazzite image** (`rpm-ostree install unzip` required), so install will fail with `unzip: command not found` until layered. Same story for `vulkan-tools`. wl-copy is available on KDE Wayland.
- **Nobara:** ✅ likely — `unzip` and `vulkan-tools` typically preinstalled on this gaming-tooling distro.
- **ChimeraOS:** ✅ likely — gaming-handheld image with `unzip` + Vulkan tooling preinstalled.

- **`unzip` is in `permissions.commands` but the backend doesn't detect its absence** (backend.ts:415). Unlike `vulkaninfo`/`vkcube`/`wl-copy`/`xclip`, the install path goes straight to `run(["unzip", …])` and surfaces `unzip failed (exit ${exitCode}): ${stdout}` — but a missing-binary case wouldn't have a meaningful exit/stdout, it would `ENOENT`-throw. On Bazzite this produces a confusing error rather than a "install unzip" hint.
- **`STEAM_DEFAULT_DLL()` falls through to a non-existent path in `_resolveDllForToml`** (backend.ts:488). If `getDllStatus` returns `path: null` AND no custom path is set, `renderTomlConfig` will write `dll = "/home/u/.local/share/Steam/.../Lossless.dll"` even though the file doesn't exist. The TOML will then fail at lsfg-vk runtime with a layer-side error rather than a setup-time error. Not a regression vs. the upstream Decky port, just worth flagging.
- **`launchVkcube` builds a process env from `process.env`** (backend.ts:307-315). Since the loader runs as root (per the project's root-service architecture), `vkcube` spawns with root env — including a root `$HOME` and `$XDG_RUNTIME_DIR`. On SteamOS / Bazzite this may not connect to the user's Wayland socket. The UI says "Test with vkcube" but on the actual install target it may silently fail or display nothing. Flagged for the integration-test pass.
- **`_findExtractedFiles` BFS doesn't dedupe / early-exit** (backend.ts:459). If the upstream zip ever ships two copies of `liblsfg-vk.so` (e.g. arch subdirs), the second one wins. Unlikely with the current noui zip layout, but the BFS is unbounded.
- **Optimistic local launch-options map can drift** (GamePicker.tsx:179-208). Comment correctly identifies the Steam-localconfig flush race, but if `appendLaunchToken` returns the new string and Steam's autosave never flushes (e.g. user kills Steam before autosave), the next plugin remount will see the old VDF and the row will revert to "not applied" while the game itself still has the wrapper applied in Steam's in-memory state. Symptom is "Apply button reappears for a game that's actually already applied."

PR #76, opened 2026-05-29
Migrate Theme Loader plugin (theme-loader)

### LINKED ISSUE #21: Migrate plugin: Theme Loader (theme-loader)
> **Auto-detected** — source: `linux-gaming-plugin-manager/plugins/theme-loader/` · UI: overlay (app.tsx) · removed-package deps to inline: none · subprocess: .

# Migrate one plugin: Theme Loader (`theme-loader`)

**Goal:** Port the `theme-loader` plugin from the old Steam Loader repo into the new Loadout repo, one plugin at a time. Port it *faithfully* (zero behavior regressions), simplify only where it is provably safe, and keep all of its code *inside the plugin* unless something is genuinely shared by 2+ already-migrated plugins.

## Locations
  - This plugin: `plugins/theme-loader/`
  - Port into: `plugins/theme-loader/`

---

**Layout** — `plugins/theme-loader/`:
- `package.json` — name `@loadout/plugin-theme-loader`, `"type": "module"`, deps, and a `plugin` field (the manifest now lives here, NOT in a separate `plugin.json`):
  ```json
  {
    "name": "@loadout/plugin-theme-loader",
    "version": "0.0.1",
    "type": "module",
    "plugin": {
      "id": "theme-loader",
      "name": "Theme Loader",
      "description": "…",
      "category": "…",
    }
  }
  ```

```bash
```

---

| Distro | Notes |
|---|---|

**Binary classification:**
  - `hyprctl` — Hyprland only.

---

## Porting procedure

1. **Copy** the source plugin tree `plugins/theme-loader/` from SOURCE into TARGET `plugins/theme-loader/`.
3. **Fold `plugin.json` into `package.json`.** Move the manifest fields into the `plugin` field of `package.json`, set `name` to `@loadout/plugin-theme-loader`, add `"type": "module"` and the real `dependencies`. Delete the standalone `plugin.json`.

---

|---|---|---|
| `plugin-storage` | ×8 | Inline: read/write JSON at `~/.config/loadout/plugins/theme-loader.json` (~30–50 LOC). |
| `sgdb-art` | ×2 | Inline. |

---

**Safe to cut / collapse:**

**NOT safe — do NOT touch:**

---

## Definition of Done

- [ ] `bun run typecheck`
- [ ] `bun run build`.
- [ ] **Plugin loads**: the loader logs `Loaded plugin: Theme Loader (theme-loader) …` (from `apps/loadout/src/loader/plugin-manager.ts`).

---

- **Plugin id:** `theme-loader`
- **Plugin name:** `Theme Loader`
- **Source path:** `linux-gaming-plugin-manager/plugins/theme-loader/`

srsholmes:
Closes #21.

## Summary

Faithful port of the Theme Loader plugin from the `linux-gaming-plugin-manager` repo (`@steam-loader/*`) into Loadout (`@loadout/*`). Browses, installs, configures, and toggles community CSS theme packs from deckthemes.com against Steam's Big Picture UI via Chrome DevTools Protocol — no behaviour regressions, no new RPC names/signatures.

LOC delta: 3348 (source) → 3312 (target) — within 1%, no functional simplifications.

## Shared packages consumed

- `@loadout/ui` — `mountComponent`, `mountHeaderStub`, `PluginProvider`, `PluginHeader`, `useBackend`, `useFocusable`, `useIntersectionGate`, `Badge`, `Button`, `HeaderBackButton`, `IconButton`, `Panel`, `SearchField`, `SegmentedItem`, `Select`, `Spinner`, `Text`, `Toggle`, `notify`.
- `@loadout/types` — `PluginBackend`, `EmitPayload`.
- `@loadout/exec` — `runCode` (xdg-open), `runFull` (unzip).
- `@loadout/steam-cdp` — `CDPClient`. The multi-tab injection requires holding several CDP connections open simultaneously (one per Steam target tab) so `withSteamClient` isn't a fit; the raw `CDPClient` is what the source plugin uses and what's preserved here.
- `@loadout/plugin-storage` — `readPluginStorage` / `writePluginStorage`. Collapses the source's two on-disk JSON files (`active-themes.json` + `pack-variants.json` under `~/.config/steam-loader/theme-loader/`) into the one-file-per-plugin contract at `~/.config/loadout/plugins/theme-loader.json`.

## Target

- `target.type: "overlay"` — the settings + community browser UI renders inside the Electrobun overlay (`app.tsx`).
- Theme CSS is injected into Steam's CEF surfaces from the backend via CDP. **Multi-target injection is preserved verbatim** — connections held open against `SharedJSContext` / `Steam` / `SP` / `Steam Big Picture Mode` / `MainMenu*` (prefix) / `QuickAccess*` (prefix). The `MainMenu_uid2` popup quirk noted in the project memory (`css_loader_bpm_tabs`) is intact: missing this target makes BPM themes silently no-op visually.

## Permissions

- `permissions.network`: `api.deckthemes.com`, `share.deckthemes.com`, `cdn.deckthemes.com`, `github.com`, `codeload.github.com`, `localhost` (loopback for CEF `/json` tab discovery).
- `permissions.commands`: `unzip`, `xdg-open`.
- `permissions.filesystem`: `~/.local/share/loadout/theme-loader/css-themes`, `~/.cache/loadout/theme-loader`.

## Per-distro verdict

| Distro | Verdict | Notes |
|---|---|---|
| SteamOS (stock Deck) | ✅ likely | Steam CEF + `unzip` + `xdg-open` all present. `XDG_CONFIG_HOME` aware. |
| CachyOS | ✅ likely | All commands present; `xdg-open` standard via xdg-utils. |
| Bazzite | ✅ likely | Same as above; storage paths fall under `$XDG_*` which works under rpm-ostree. |
| Nobara | ✅ likely | Fedora gaming distro; `unzip` and `xdg-open` preinstalled. |
| ChimeraOS | ✅ likely | Steam Gaming Mode + Arch base. |

No bundled binaries; everything declared in `permissions.commands` is universal across the matrix.

## Deviations from source

- **Storage layout**: source persisted state to two custom files (`~/.config/steam-loader/theme-loader/active-themes.json` + `pack-variants.json`). Target uses `@loadout/plugin-storage` which keys one JSON file per plugin under `~/.config/loadout/plugins/theme-loader.json`. Same on-disk semantics (atomic writes, JSON shape), one file instead of two. No observable behaviour change beyond path.
- **Cache env var rename**: `STEAM_LOADER_CSS_CACHE_DIR` → `LOADOUT_THEME_CSS_CACHE_DIR`, `STEAM_LOADER_THEME_CACHE_DIR` → `LOADOUT_THEME_CACHE_DIR`. Same purpose, scoped to the new product name.
- **Data dir paths** moved from `~/.local/share/steam-loader/css-themes/` and `~/.cache/steam-loader/theme-loader/` to `~/.local/share/loadout/theme-loader/css-themes/` and `~/.cache/loadout/theme-loader/`. Existing installs on a Loadout-only system have nothing under the old path; no migration shim is necessary per the project's "no legacy support" rule.
- **Inline `<style>` dataset attr** renamed from `data-steamLoaderPlugin="theme-loader"` to `data-loadoutPlugin="theme-loader"`. Diagnostic-only; not referenced by the runtime.
- **Tests**: source ran `app.spec.tsx` under vitest; target rewrites it on top of `bun:test` + `mock.module` + the captured-real-module pattern documented in `docs/test-mock-contamination.md`.

## CI gates

```
bun run typecheck       PASS
bun run lint            0 errors, 35 warnings (= baseline)
bun run check:specs     OK
bun run test:backend    942 pass, 0 fail
bun run test:ui         168 pass, 0 fail
bun run build           OK
```

## Test plan

- [ ] `bun run dev:overlay` boots; Theme Loader appears in the sidebar with the palette icon.
- [ ] Community tab loads from `api.deckthemes.com`; sort options reorder; search filters by name/author/description.
- [ ] Install a theme → it appears under Themes with a thumbnail + author + license label.
- [ ] Toggle the theme on while Steam BPM is running → visible CSS applied to BPM (MainMenu popup) + SharedJSContext + QuickAccess.
- [ ] Toggle off → CSS removed cleanly from all tabs.
- [ ] Theme with patches: expand row, change a variant, see live re-injection.
- [ ] Reapply icon button reconnects to Steam CEF after a Steam restart.
- [ ] Uninstall removes the directory + persisted variants; the theme disappears from the Themes tab.
- [ ] Survive a service restart: previously-active themes are re-injected on next `tryConnect()`.


srsholmes:
## PR #76 — theme-loader migration review

### Code quality
- **`tee` declared but unused** (`package.json:28`): `permissions.commands` lists `["unzip", "xdg-open"]`. The plain text says `tee` — actually it doesn't, my mistake. The list is exact: only `unzip` (used at `backend.ts:410`) and `xdg-open` (used at `backend.ts:548`). Clean.
- **Unused/over-declared network hosts** (`package.json:16-23`): `share.deckthemes.com`, `cdn.deckthemes.com`, `github.com`, `codeload.github.com` are declared, but the backend only fetches `api.deckthemes.com` (`backend.ts:380, 587` + `themes-cache.ts:150` + `translations-cache.ts:26, 86`). Thumbnails resolve to `${API_BASE}/blobs/{id}` (`themes-cache.ts:98`), not the CDN. The sandboxed-fetch is deny-by-default, so the extras are inert — but flag as cleanup. (`localhost` is needed for CEF debug port at `:8080`.)
- **`console.log` left in production paths** (`backend.ts:134, 153, 158, 183, 379, 596, 612, 700, 733`; `themes-cache.ts:190`; `translations-cache.ts:117`). These match source-plugin style and the project's other plugins use the same `[plugin-id]` prefix convention — not a regression, noted only because the migration prompt calls vestigial `console.log` as a target. These look intentional.
- **`@loadout/steam-cdp` import** (`backend.ts:3`): correctly uses the new shared package per the migration spec. No raw CDP impl inlined.
- **Mount boilerplate**: `mountComponent(ThemeLoader)` + `mountHeaderStub` (`app.tsx:855-856`) — matches the convention exactly.
- **Cross-plugin / disallowed imports**: none. Only `@loadout/{ui,types,exec,steam-paths,plugin-storage,steam-cdp}`, react, react-icons, and node built-ins.
- **No `Bun.spawn`, no `sudo`/`pkexec`** in any file. Subprocess routes through `@loadout/exec` (`runFull`, `runCode`).
- **`writeThemeMeta` dynamic import** (`theme-pack.ts:129`): `const { writeFile } = await import("node:fs/promises")` while the rest of the file uses a top-level `import { …, readFile, … } from "node:fs/promises"`. Should be hoisted to the static import — defer-load buys nothing here. Minor.
- **Inline duplicate of `Bun.file().exists()` semantics**: `locateThemeRoot` (`theme-pack.ts:153-173`) does two-level recursive readdir-and-probe. Defensible — manifest_v2 zips have variable layouts. Inline is correct per the spec; no shared helper.
- **Pure helpers inside `theme-pack.ts`** already split out: `findUpstreamLicense`, `locateThemeRoot`, `collectInjectFiles`, `assemblePackCss`, `summarizePatches`, `selectDefaultVariant`, `translateCss`. Good decomposition. `selectDefaultVariant` is module-private but tested transitively via `collectInjectFiles` — borderline; would be cleaner exported with a direct unit test.
- **Pure helpers still in `backend.ts`**: `isTargetTab` (line 102) and `safeStyleId` (line 113) are module-scope pure functions. They'd be testable in isolation as `lib/tab-matching.ts`. Light flag — both are <10 LOC and tightly coupled to backend domain.

### Bundled-binary quality
No `bundled_bins` declared. `unzip` and `xdg-open` are universal user-space utilities — correct call not to bundle.

### Lint-warnings count
Baseline (main): **44** — PR: **44** — Delta: **0**. (Note: baseline drifted from the 35 the spec mentions; main is the source of truth.) Zero new errors. Zero new `as any` in plugin source; one `as unknown as` cast in `backend.test.ts:64` (`globalThis.fetch = mock(...) as unknown as typeof fetch`) — standard test-mock pattern.

- `bun run check:specs`: **OK** — every ≥100 LOC entry has a sibling spec.
- `bun run test:backend`: **1353 pass / 0 fail**.
- `bun run test:ui`: **253 pass / 0 fail**.
- Coverage gaps:
  - **`assemblePackCss` has no direct test** (`theme-pack.ts:253`). It's the canonical CSS-assembly + translation path; only `findUpstreamLicense` and the meta sidecar round-trip are tested in `theme-pack.test.ts`. Would be straightforward unit-testable (synthetic manifest + tmpdir + assert string contains `:root { --foo: bar; }` + translated-class swap). Real risk: silent regression in CSS variable injection or longest-first translation ordering.
  - **`collectInjectFiles` has no direct test** (`theme-pack.ts:212`) — handles patch/variant fallback semantics that determine which CSS files are picked. Untested.
  - **`translateCss` has no direct test** (`theme-pack.ts:25`) — word-boundary regex correctness is non-trivial.
  - **`isTargetTab` has no direct test** (`backend.ts:102`) — single most important matcher in the plugin (memory-flagged BPM popup discovery hinges on it). If `MainMenu` prefix ever needs to be tighter, there's no test to catch a regression.
  - **`buildMap` translation construction** (`translations-cache.ts:57`) is tested transitively via `ensureTranslations` — fine.

|---|---|---|
| `unzip` | ✅ Universal | Preinstalled on SteamOS, CachyOS, Bazzite, Nobara, ChimeraOS. |
| `xdg-open` | ✅ Universal | `xdg-utils` shipped by default everywhere. |
| `localhost:8080` (CEF debug) | ⚠️ Gamescope-conditional | Requires Steam launched with `--cef-enable-debugging`. SteamOS Gaming Mode and Bazzite-Deck Gaming Mode pass this by default. Desktop-mode Steam needs the flag set manually — plugin no-ops gracefully via `tryConnect()` catch (`backend.ts:621`). |
| `~/.local/share/loadout/theme-loader/css-themes` | ✅ Universal | XDG home. |
| `~/.cache/loadout/theme-loader` | ✅ Universal | XDG cache. |
| Network: `api.deckthemes.com` | ✅ Universal | Public HTTPS API. |

**Verdict (out of the box):**
- SteamOS (stock Deck): ✅ likely — full Gaming Mode CEF debug + universal tools.
- CachyOS:              ✅ likely on Gaming Mode / gamescope sessions; ⚠️ partial in desktop Steam without the debug flag.
- Bazzite:              ✅ likely on Bazzite-Deck / Gaming Mode; ⚠️ partial on Bazzite Desktop without `--cef-enable-debugging`.
- Nobara:               ⚠️ partial — desktop Steam users need to enable CEF debugging.
- ChimeraOS:            ✅ likely — handheld Gaming Mode default.

- **Update destroys previous install on failure** (`backend.ts:425-428`): `installCommunityTheme` removes `targetDir` BEFORE the new `cp` runs. If `cp(themeRoot, targetDir)` throws (disk full, EIO, ENOSPC), the user's previously-installed theme is gone and the error message says "Install failed:" with no recovery hint. A staged install (cp to `targetDir.tmp`, then atomic rename) would survive this.
- **`MainMenu` prefix matches too greedily** (`backend.ts:97`): `tab.title.startsWith("MainMenu")` matches any future tab with that prefix (e.g. `MainMenuOptions` or a popup Valve adds in a Steam update). For BPM popup it's correct today, but the source plugin had the same pattern — flagging because the memory note (`project_css_loader_bpm_tabs.md`) is the reason this is delicate. If Valve renames the popup, themes silently no-op visually — exactly the failure mode the memory warns about.
- **Health-check reconnect can double-inject** (`backend.ts:706-714`): `checkHealth` runs every 5s. If `tryConnect()` succeeds AND the user *had* an active theme that `tryConnect()` already re-injected at lines 148-155, `checkHealth` would re-inject again via the new `for ... of activeThemes` loop. `injectCSSToTab` removes the existing `<style id>` first (`backend.ts:641`), so this is idempotent in practice — but it's a wasted round-trip every 5s while disconnected-then-reconnected. Real bug only if injectCSS is non-idempotent for any future code path.
- **`tryConnect()` resets `this.connections` without closing failed sockets** (`backend.ts:602-616`): the loop closes existing connections, then `for (const tab of targetTabs)` opens new ones via `openCDP`. If `openCDP` resolves but later throws between resolution and `this.connections.push`, the socket leaks. Currently `openCDP` is a thin wrapper — unlikely to leak, but the existing-connection-close happens unconditionally even if no new tabs ultimately attach (e.g. all `openCDP` calls fail), leaving the plugin in a "no connections, was just closed" state on a transient registry blip.
- **`emit?.({ event: "themesChanged", … })` not emitted on `setThemePackVariant` storage failure** (`backend.ts:521-535`): variant changes call `saveStateToDisk()` (line 523) which silently swallows fs errors (`backend.ts:746-748`). On disk failure the in-memory state still updates and the UI emits `themesChanged`, but the persisted state is now divergent. Minor — symptom is "variant reverts on restart".
- **`writeThemeMeta` dynamic `import("node:fs/promises")`** (`theme-pack.ts:129`): functional but inconsistent with rest of file. Could fail in a bundler that doesn't handle dynamic imports of node built-ins — currently fine under Bun.
- **`fetch(`http://localhost:${DEBUG_PORT}/json`)` not declared explicitly** (`backend.ts:587`): `localhost` IS in the network list, so this passes sandbox. Sanity-check only.

PR #77, opened 2026-05-29
Migrate Sound Loader plugin (sound-loader)

### LINKED ISSUE #24: Migrate plugin: Sound Loader (sound-loader)
> **Auto-detected** — source: `linux-gaming-plugin-manager/plugins/sound-loader/` · UI: overlay (app.tsx) · removed-package deps to inline: none · subprocess: .

# Migrate one plugin: Sound Loader (`sound-loader`)

**Goal:** Port the `sound-loader` plugin from the old Steam Loader repo into the new Loadout repo, one plugin at a time. Port it *faithfully* (zero behavior regressions), simplify only where it is provably safe, and keep all of its code *inside the plugin* unless something is genuinely shared by 2+ already-migrated plugins.

## Locations
  - This plugin: `plugins/sound-loader/`
  - Port into: `plugins/sound-loader/`

---

**Layout** — `plugins/sound-loader/`:
- `package.json` — name `@loadout/plugin-sound-loader`, `"type": "module"`, deps, and a `plugin` field (the manifest now lives here, NOT in a separate `plugin.json`):
  ```json
  {
    "name": "@loadout/plugin-sound-loader",
    "version": "0.0.1",
    "type": "module",
    "plugin": {
      "id": "sound-loader",
      "name": "Sound Loader",
      "description": "…",
      "category": "…",
    }
  }
  ```

```bash
```

---

| Distro | Notes |
|---|---|

**Binary classification:**
  - `hyprctl` — Hyprland only.

---

## Porting procedure

1. **Copy** the source plugin tree `plugins/sound-loader/` from SOURCE into TARGET `plugins/sound-loader/`.
3. **Fold `plugin.json` into `package.json`.** Move the manifest fields into the `plugin` field of `package.json`, set `name` to `@loadout/plugin-sound-loader`, add `"type": "module"` and the real `dependencies`. Delete the standalone `plugin.json`.

---

|---|---|---|
| `plugin-storage` | ×8 | Inline: read/write JSON at `~/.config/loadout/plugins/sound-loader.json` (~30–50 LOC). |
| `sgdb-art` | ×2 | Inline. |

---

**Safe to cut / collapse:**

**NOT safe — do NOT touch:**

---

## Definition of Done

- [ ] `bun run typecheck`
- [ ] `bun run build`.
- [ ] **Plugin loads**: the loader logs `Loaded plugin: Sound Loader (sound-loader) …` (from `apps/loadout/src/loader/plugin-manager.ts`).

---

- **Plugin id:** `sound-loader`
- **Plugin name:** `Sound Loader`
- **Source path:** `linux-gaming-plugin-manager/plugins/sound-loader/`

srsholmes:
Closes #24

## Summary

Ports the **Sound Loader** plugin (custom Steam UI sound packs from deckthemes.com) from steam-loader to Loadout.

Faithful port — same RPC method names, event names, payload shapes, and the load-bearing `PlayAudioURL`/webpack-source-search injection that swaps Steam's UI sounds at runtime.

## Shared packages consumed

- `@loadout/ui` — `mountComponent`, `mountHeaderStub`, `PluginHeader`, `useBackend`, `useFocusable`, plus the component library (`Badge`, `Button`, `HeaderBackButton`, `SearchField`, `SegmentedItem`, `Spinner`, `Toggle`, `notify`).
- `@loadout/types` — `PluginBackend`, `EmitPayload`.
- `@loadout/exec` — `runFull` for the community-pack `unzip` step.
- `@loadout/plugin-storage` — `readPluginStorage` / `writePluginStorage` (replaces the hand-rolled `~/.local/share/steam-loader/sound-loader.json`).
- `@loadout/steam-cdp` — `CDPClient` + `findSharedJsTab` (replaces `@steam-loader/injector`'s `findSharedJSContext`).

Plus `react`, `react-dom`, `react-icons`.

## Target

- `target` omitted → defaults to overlay (the only surface). Backend talks to Steam's CEF via `@loadout/steam-cdp`.
- `loadOnStartup: true` so `init()` applies persisted overlay overrides before the user opens the UI.

## PlayAudioURL / injector mechanism — preserved exactly

The load-bearing injection sequence is unchanged from source:

1. `findSharedJsTab` (one-shot variant from `@loadout/steam-cdp`) locates the SharedJSContext tab — returns `null` when Steam isn't reachable; the source's `findSharedJSContext` threw, so `tryConnect()` now handles the null case explicitly.
2. Open a CDP WebSocket against the tab's `webSocketDebuggerUrl`.
3. Wait for `webpackChunksteamui` (poll every 100ms, 5s ceiling).
4. Inject `STEAM_HOOK_SCRIPT` from `lib/steam-hook.ts` — pushes a probe chunk to capture `req`, source-searches `req.m` for the `PlayAudioURL(<arg>){` signature with the `PlayAudioURLWithRepeats|m_Context|destination` co-occurrence filter, force-`req()`s each candidate, finds the `AudioPlaybackManager` class via `prototype.PlayAudioURL`, wraps the prototype method with a `basename(url)` lookup into `window.__SL_AUDIO_OVERRIDES__`.
5. Stage active pack's WAVs into `~/.local/share/Steam/steamui/sounds_custom/loadout/` and push the canonical Decky-filename → `https://steamloopback.host/...` URL map via `__SL_AUDIO_OVERRIDES__`.
6. Health-check loop reconnects on WebSocket drops; `Page.frameNavigated` re-injects after Steam reloads.

Decky-symlink reclaim logic (`reclaimStagingParent`) preserved verbatim — broken `sounds_custom` symlinks from Decky uninstall are detected via `lstat` and replaced with a real directory.

## File paths declared

- `permissions.filesystem`: `~/.local/share/Steam/steamui/sounds_custom/loadout` (Steam-side staging dir for sound-file swaps).
- Pack storage (`~/.local/share/loadout/sound-packs/`) and cache (`~/.cache/loadout/sound-loader/`) live under the user's data dirs — the loader's `fs`-write gap covers these without an explicit declaration today (same as the lsfg-vk reference).

## Network

`permissions.network`: `api.deckthemes.com`, `localhost`, `127.0.0.1`.

- `api.deckthemes.com` — community-pack registry + blob CDN.
- `localhost` / `127.0.0.1` — Steam's CEF debug port (`8080`) for the CDP injection.

## Commands

`permissions.commands`: `["unzip"]` — used to extract community-pack blob zips. Universal across SteamOS / CachyOS / Bazzite / Nobara / ChimeraOS.

## Other notes

- Storage namespace migrated: `~/.local/share/steam-loader/sound-packs/` → `~/.local/share/loadout/sound-packs/`; `~/.cache/steam-loader/sound-loader/` → `~/.cache/loadout/sound-loader/`; `sounds_custom/steam-loader/` → `sounds_custom/loadout/`. Per project convention (no legacy support; sole user).
- Backend paths resolve lazily from `process.env.HOME` so tests sandbox via env override (lsfg-vk + fan-control pattern); old `_testOverridePaths` dead-code dropped.
- Tests converted vitest → `bun:test`: `backend.test.ts` + `lib/*.test.ts` under bun's native env, `app.spec.tsx` under happy-dom via `mock.module("@loadout/ui", ...)` with a static `import * as actualUi` so `PluginProvider` can be shimmed to `PluginHeaderSlotProvider`. 33 backend/lib tests + 11 UI tests pass.
- Mount entry points use `mountComponent` + `mountHeaderStub`; dynamic header still portals through `<PluginHeader>` inside the body tree so `activeTab` + `communityQuery` stay closure-shared (no cross-root pub/sub).
- LOC delta vs source: **−13** (essentially flat — `mountComponent` + `plugin-storage` replace inline `createRoot + PluginProvider` and config helpers).

## Per-distro verdict

Overlay-only plugin; depends on `unzip` + Steam's CDP debug port being reachable.

| Distro | Verdict | Reason |
|---|---|---|
| SteamOS | ✅ likely | `unzip` present in stock image; Gaming Mode runs Steam with the debug port open. |
| CachyOS | ✅ likely | `unzip` standard package; works on any session type. |
| Bazzite | ✅ likely | `unzip` ships in the rpm-ostree base; KDE/SteamOS variants both fine. |
| Nobara | ✅ likely | Same Fedora base; `unzip` present. |
| ChimeraOS | ✅ likely | Arch-based handheld image; `unzip` present. |

## Lint baseline

35 → 37 (delta **+2**, within the cap). Both new warnings are the same `any` casts on `PluginProvider` / `useEvent` mocks used by the fan-control reference spec.

## Test plan

- [x] `bun run typecheck`
- [x] `bun run lint` (37 warnings; no new errors; +2 vs main = 35)
- [x] `bun run check:specs` (≥100 LOC files have sibling specs)
- [x] `bun run test:backend` (946 / 946 pass)
- [x] `bun run test:ui` (171 / 171 pass)
- [x] `bun run build`
- [ ] Smoke: load Loadout, open Sound Loader plugin, browse community packs, install one, toggle "Steam UI sounds" with Steam running, verify Steam's UI sounds get replaced.


srsholmes:
## Code review — PR #77 (sound-loader)

Independent correctness/performance/security pass complementing the migration review. Plugin is well-structured, follows the established patterns (lazy `HOME`, `bun:test`, `mountComponent` + `<PluginHeader>` portal, `@loadout/plugin-storage`), and ships strong test coverage (33 backend/lib + 11 UI). Findings below are graded by severity.

### Correctness

**[medium] `_applySteamState` is not serialized against itself.**
`backend.ts` fires `void this._applySteamState()` from `setActivePack`, `setUseInSteam`, `uninstallCommunityPack`, the health-check reinject hook (`startMonitor(() => this._applySteamState())`), and `reconnectSteam`. Two concurrent invocations will interleave `clearStagedFiles(stagingDir)` → `prepareStagingDir` → `cp(...)` against the same `~/.local/share/Steam/steamui/sounds_custom/loadout/` tree. If a user toggles "Steam UI sounds" while Steam reloads (`Page.frameNavigated`), invocation B's `clearStagedFiles` can wipe invocation A's freshly-staged files between A's `stagePackFiles` returning and A's `refreshOverrides` evaluating — Steam ends up with an override map pointing at deleted files. Suggest a simple `Promise` mutex (the TDP debounce pattern in this codebase already does this for SMU writes).

**[medium] `reconnectSteam` swaps `this.steamInjector` while `_applySteamState` may be holding a stale reference.**
`_applySteamState` captures `const injector = this.steamInjector` at entry, then `await`s on `injector.tryConnect()`, `injector.injectHook()`, `stagePackFiles(...)`, `injector.refreshOverrides(...)`. If `reconnectSteam` runs during one of those awaits, it `await`s `this.steamInjector.stop()` (closes the OLD CDP socket) and assigns a new instance — but the in-flight `_applySteamState` keeps driving the closed injector. Result: a `not connected` error event, plus the new injector gets `startMonitor`'d twice once the new `_applySteamState` lands. Either guard with a generation counter or await the in-flight apply in `reconnectSteam`.

**[low] `installCommunityPack` extracts attacker-controlled zip without `-:` (no-extraction-outside-target) flag.**
`runFull(["unzip", "-o", zipPath, "-d", extractDir])` allows entries with `../` in their names to escape `extractDir` on `unzip` builds without `-:` enabled by default. The `extractDir` is in `tmpdir()`, so the blast radius is "anywhere the user can write" rather than this plugin's namespace, but with a malicious `deckthemes.com` response an attacker could drop files anywhere on the filesystem. Add `-:` (or post-validate that every extracted path resolves under `extractDir`). The downstream `_locatePackRoot` + `readdir` + `cp` to `SOUND_PACKS_DIR/id` does NOT re-validate either, so the escaped payload would survive into the install dir if `_locatePackRoot` happens to surface it.

**[low] `uninstallCommunityPack` traversal guard misses the base directory.**
`!resolvedTarget.startsWith(resolvedBase + "/")` returns `true` for `resolvedTarget === resolvedBase`. `SAFE_ID` already blocks empty strings and `..`, so this is defense-in-depth, but the cleaner check is `dirname(resolvedTarget) === resolvedBase`.

**[low] `init()` discards subscription handles — handler leak on hot reload.**
`app.tsx:1098-1099` calls `api.subscribe(...)` twice and never stores the returned unsubscribe fns. If the loader ever re-runs `init()` (hot reload, plugin disable/enable), each reload accumulates a new pair of `reapply` callbacks on the same events.

**[low] `_loadConfig` swallows malformed-JSON without surfacing.**
If `~/.local/share/loadout/plugin-storage/sound-loader.json` gets corrupted, the bare `catch {}` silently resets to defaults. A `console.warn` would match the pattern used in `_scanPacks` and `_saveConfig`.

**[nit] Unused-by-runtime constants.** `SOUND_EVENTS` is defined identically in `backend.ts` and `app.tsx` — diverging definitions would silently break the event-mapping flow. Worth a small shared module under `lib/` (or `@loadout/types`) since both ends are bundled in the same repo.

### Performance

**[low] `installSoundOverrides` re-decodes every base64 buffer on every config change.**
`app.tsx:721-727` re-runs `installSoundOverrides` whenever `activePack | useInOverlay | loading` change. `getActivePackMappings` returns full base64 audio, `decodeBase64Audio` runs an `atob` + `decodeAudioData` per file. For a multi-file pack this can be several MB of decode per toggle. Cache `AudioBuffer`s keyed by `(packId, event, fileIndex)` and invalidate only when `packId` actually changes.

**[low] `listCommunityPacks` does a fresh `readdir(SOUND_PACKS_DIR())` on every call.**
Cheap, but it's called every time the Community tab re-renders or `onPackInstalled` fires. Cache invalidation already runs through `_scanPacks` — reusing `this.packsCache.keys()` as the installed-set source would avoid the extra fs hit and stay consistent with the rest of the backend.

**[low] Pagination polls 50ms after the last full page in `sounds-cache.fetchUpstream`.**
Not a real issue (legacy endpoint terminates fast), but `MAX_PAGES = 50 * PER_PAGE = 2500` worth of paging with 30s timeouts each means a misbehaving server can stall `ensureCommunityPacks` for ~25 minutes before giving up. Consider a wall-clock budget on the whole pagination loop in addition to per-request `FETCH_TIMEOUT_MS`.

**[low] AudioContext never closes.**
`audioCtx` in `app.tsx:110` is module-level; if the plugin unmounts (overlay closes), the context leaks. `mountComponent`'s unmount hook would be the place to `audioCtx.close()`.

### Security

**[ok] `STEAM_HOOK_SCRIPT` is a static string template — no untrusted interpolation.** Override values arrive separately via `JSON.stringify(map)` evaluated as the RHS of an assignment. Map values are URLs we control (`https://steamloopback.host/...` with `encodeURIComponent` on filenames). Clean.

**[ok] `SAFE_ID` regex blocks path traversal and shell metachars for pack ids before they reach `join(SOUND_PACKS_DIR(), id)`.**

**[ok] `_readAudioFile` traversal check rejects `..`, `/`, `\\`.**

**[medium] Network allowlist is broader than necessary.**
`permissions.network` includes `localhost` AND `127.0.0.1` — the CDP injector hard-codes `localhost:8080` via `findSharedJsTab` (`@loadout/steam-cdp`). If the sandbox resolves the allowlist by string match, declaring both is harmless; if it does DNS resolution, `localhost` is sufficient. Trim if possible to minimize the attack surface.

**[low] `installCommunityPack` trusts `pack.downloadUrl` from the registry without re-pinning to `api.deckthemes.com`.**
The PR body says "Always uses `api.deckthemes.com/blobs/<id>`", but the code does `fetch(pack.downloadUrl, ...)` with whatever the upstream JSON returned. If the upstream JSON ever returns a non-deckthemes URL (registry compromise, MITM on the registry response — registry IS HTTPS so this is "upstream malice" tier), the install proceeds against it. Hardening: assert `new URL(pack.downloadUrl).hostname === "api.deckthemes.com"` before fetch.

**[low] `unzip` extraction without entry-path validation (see correctness section above).**

### Test coverage

**[ok] Backend tests** cover config persistence, RPC event emission, registry status state machine, cache file dedupe, traversal rejection.

**[ok] `lib/steam-injector.test.ts`** covers the Decky symlink reclaim edge cases (the load-bearing reason this file exists) and stale-file cleanup.

**[gap] `_applySteamState` has no integration coverage.** The full flow — connect → injectHook → stagePackFiles → refreshOverrides → error events — is the largest surface area in the plugin and the most stateful. A test that fakes a `CDPClient` and asserts the `steamError` event lifecycle (null on success, error string on each failure mode, cleared on subsequent success) would catch the concurrent-call race above.

**[gap] `installCommunityPack` happy path is untested.** Only the two rejection paths (unknown id, traversal id) are exercised. The Decky-`pack.json`-aware mapping logic (`backend.ts:478-544`) is non-trivial and would benefit from a fixture-driven test.

**[gap] `app.spec.tsx` does not exercise the Community tab.** The community grid, search filter, install/uninstall buttons, and refresh flow all live in `CommunityPacks` and are not touched by the spec. The error-state retry surface is also untested.

### Summary

Three correctness items worth addressing before merge: the unsynchronized `_applySteamState`, the reconnect-vs-apply race, and the unzip extraction guard. Everything else is incremental hardening that can ship in a follow-up. The injection mechanism preservation and the deckthemes-leftover symlink reclaim logic are well-tested and the load-bearing pieces look correct.

srsholmes:
## Migration review — PR #77 (sound-loader)

Branch is 1 commit behind `main` (only #78 / steamgriddb merged after this branched) — no rebase needed.

### Code quality
- `backend.ts:628-631` — dead branch: `readdir` returns an array (or throws); `if (!entries)` is unreachable. The `try/catch` already handles the not-installed case.
- `backend.ts:150-159` and `:251-257` — duplicated `AudioSteamInjector` logger factory (same 4-line closure). Could be a `private makeLogger()` helper.
- `backend.ts:96-98` — local `CommunityPackInfo extends CommunityPackEntry` declared in backend AND re-declared standalone in `app.tsx:57-70`. Both halves drift independently. Same shape, two sources of truth.
- `app.tsx:32-43` — `SOUND_EVENTS` const literal duplicated verbatim from `backend.ts:26-37`. Same for `EVENT_TO_METHOD` (not in backend, but the event keys are). One inline copy is the migration-prompt rule until 2+ consumers; sound-loader is the only consumer today, so fine — flag only if/when shared.
- `app.tsx:559-1048` (~490 LOC of `SoundLoader()`) mixes header JSX, body JSX, two tabs, and 4 callbacks. The `CommunityPacks` subcomponent is already extracted in-file (`:276-557`); pulling it into `lib/community-packs.tsx` would drop ~280 LOC from `app.tsx`. Not blocking — the migration prompt sanctions decomposition when it crosses the readability threshold; subjective call.
- Imports clean: only `@loadout/{ui,types,exec,plugin-storage,steam-cdp}` + `react*` outside plugin files. No `Bun.spawn`, no `sudo`/`pkexec`, no `@steam-loader/*` remnants, no cross-plugin imports. `mountComponent` + `mountHeaderStub` used correctly.
- `sounds-cache.ts:255-264` — `_resetForTests` exported from production module (test-only). Tolerable; same pattern as themes-cache, per comment.

### Bundled-binary quality
No `bundled_bins` declared — plugin uses system `unzip` only. `unzip` is preinstalled on all 5 target distros, so no bundling needed.

### Lint-warnings count
Merge-base `82bb6e1` (main): 35 — PR: 37 — **Delta: +2**
- Both new warnings are in `app.spec.tsx:23,30` (`@typescript-eslint/no-explicit-any` on the `PluginHeaderSlotProvider` cast and the `useEvent` handler signature in the `mock.module` block). 0 new errors. Matches the pattern in `tdp-control`/`steam-gamescope-ipc` specs — bun's mock-module typing forces these.
- Zero `as any` in production code (backend/app/lib). All `any` use is test-scoped.

- `check:specs`: ✅ pass.
- Full `bun run test:backend` after the PR: **946 pass / 0 fail** — sound-loader's 16 backend + sounds-cache + steam-injector specs all green. `test:ui` glob matches `app.spec.tsx` only — runs under happy-dom (passes there).
- Pure logic in `backend.ts` worth extracting:
  - `_locatePackRoot` (`:770-794`) — pure filesystem-search; would become `lib/locate-pack-root.ts`. 25 LOC, no `this`, mockable via tmpdir.
  - The Decky→Loadout mapping reconciliation inside `installCommunityPack` (`:478-544`, ~65 LOC of pure logic mixing `audioFiles`, `deckyMappings`, and `DECKY_TO_LOADOUT` into a `mappings` object) — would become `lib/decky-mapping.ts`. Currently untested in isolation; tested only through the full install path which requires fixture zips on disk.
  - `summaryToEntry` + `parseGithub` (`sounds-cache.ts:71-94`) — already lib-local, but no tests target them directly. Existing `sounds-cache.test.ts` covers them transitively via `ensureCommunityPacks`.

| Binary / path / domain | Class | Notes |
|---|---|---|
| `unzip` | ✅ | Preinstalled on SteamOS holo-arch base + Bazzite/Nobara Fedora base + CachyOS + ChimeraOS. |
| `~/.local/share/Steam/steamui/sounds_custom/loadout` | ⚠️ | Path only exists if Steam is installed and has rendered the UI at least once. Sound-Steam toggle is gated on `useInSteam`; UI surfaces `steamError` on connect failure, so missing dir is recoverable. |
| `api.deckthemes.com` | ✅ | Public CDN. |
| `localhost` / `127.0.0.1` | ✅ | Steam CDP at `:8080`. |
| Steam CDP `localhost:8080` (steam-cdp pkg) | ⚠️ | Only meaningful when Steam is running with `-cef-enable-debugging`. Stock Bazzite/SteamOS Steam launchers set this; flatpak Steam may not. Injector surfaces `steamError` cleanly when absent. |

**Verdict (out of the box):**
- SteamOS (stock Deck):  ✅ likely — Steam CDP available in Gaming Mode, unzip preinstalled, sounds_custom path is canonical Decky-AudioLoader location.
- CachyOS:               ✅ likely — assuming native Steam (not flatpak); flatpak Steam blocks CDP debug port.
- Bazzite:               ✅ likely — same caveat as CachyOS for flatpak Steam.
- Nobara:                ✅ likely — native Steam common; same flatpak caveat.
- ChimeraOS:             ✅ likely — Gaming Mode parity with SteamOS.

- **`backend.ts:422` — registry-controlled `downloadUrl` not host-pinned.** `fetch(pack.downloadUrl)` follows whatever the deckthemes API returns. If api.deckthemes.com is compromised or replies with a `downloadUrl` pointing at an arbitrary host, the install path proceeds (subject only to the 50 MB cap). `permissions.network` lists only `api.deckthemes.com`, but backend `fetch` isn't enforced against that list. Trust boundary extends from registry → arbitrary download host.
- **`app.tsx:491` — registry-controlled `previewImageUrl` rendered into `` with no host check.** CEF will load arbitrary URLs; hostile registry response can fingerprint users via image tracker. Same trust extension as above.
- **`sounds-cache.ts:226-237` — empty registry never reports "ready".** If upstream legitimately returns `[]`, `entries = []` (truthy but length 0), so `getCommunityPacksStatus` reports `"pending"` (or `"error"` if a prior error sticks) indefinitely. The UI's `if (status?.state === "pending" && filteredPacks.length === 0)` branch (`app.tsx:411`) will spin forever. Registry is never actually empty, so unobserved in practice.
- **`backend.ts:_applySteamState` — no serialization across concurrent callers.** Invoked from `setActivePack`, `setUseInSteam`, the injector's `startMonitor` reinject callback, `installCommunityPack` → `_scanPacks` chains, and `uninstallCommunityPack`. Two rapid toggles can interleave `clearStagedFiles` and `stagePackFiles` writes against the same dir. The `await clearStagedFiles()` → `await injector.removeOverrides()` "noOverride" branch can race a parallel staging path; staged files could outlive a disable, or get half-written. Severity is low because writes are idempotent and a 3rd call repairs state, but a single rapid toggle could leave a stale override on Steam.
- **`backend.ts:628` — dead `if (!entries)` branch (cosmetic).** See Code quality.
- **`backend.ts:_readAudioFile` is invoked from `getActivePackMappings` for every event on every overlay reapply (`app.tsx:721`).** Each call re-reads + base64-encodes the WAVs from disk. For a 10-event pack with multi-MB WAVs, every toggle / pack switch re-pays this cost. Not user-visible at handheld pack sizes today; flag for larger packs.

PR #79, opened 2026-05-29
Migrate Store Bridge plugin (store-bridge)

### LINKED ISSUE #30: Migrate plugin: Store Bridge (store-bridge)
> **Auto-detected** — source: `linux-gaming-plugin-manager/plugins/store-bridge/` · UI: overlay (app.tsx) · removed-package deps to inline: external-cache,plugin-storage sgdb-art,steam-shortcut vdf · subprocess: .

# Migrate one plugin: Store Bridge (`store-bridge`)

**Goal:** Port the `store-bridge` plugin from the old Steam Loader repo into the new Loadout repo, one plugin at a time. Port it *faithfully* (zero behavior regressions), simplify only where it is provably safe, and keep all of its code *inside the plugin* unless something is genuinely shared by 2+ already-migrated plugins.

## Locations
  - This plugin: `plugins/store-bridge/`
  - Port into: `plugins/store-bridge/`

---

**Layout** — `plugins/store-bridge/`:
- `package.json` — name `@loadout/plugin-store-bridge`, `"type": "module"`, deps, and a `plugin` field (the manifest now lives here, NOT in a separate `plugin.json`):
  ```json
  {
    "name": "@loadout/plugin-store-bridge",
    "version": "0.0.1",
    "type": "module",
    "plugin": {
      "id": "store-bridge",
      "name": "Store Bridge",
      "description": "…",
      "category": "…",
    }
  }
  ```

```bash
```

---

| Distro | Notes |
|---|---|

**Binary classification:**
  - `hyprctl` — Hyprland only.

---

## Porting procedure

1. **Copy** the source plugin tree `plugins/store-bridge/` from SOURCE into TARGET `plugins/store-bridge/`.
3. **Fold `plugin.json` into `package.json`.** Move the manifest fields into the `plugin` field of `package.json`, set `name` to `@loadout/plugin-store-bridge`, add `"type": "module"` and the real `dependencies`. Delete the standalone `plugin.json`.

---

|---|---|---|
| `plugin-storage` | ×8 | Inline: read/write JSON at `~/.config/loadout/plugins/store-bridge.json` (~30–50 LOC). |
| `sgdb-art` | ×2 | Inline. |

---

**Safe to cut / collapse:**

**NOT safe — do NOT touch:**

---

## Definition of Done

- [ ] `bun run typecheck`
- [ ] `bun run build`.
- [ ] **Plugin loads**: the loader logs `Loaded plugin: Store Bridge (store-bridge) …` (from `apps/loadout/src/loader/plugin-manager.ts`).

---

- **Plugin id:** `store-bridge`
- **Plugin name:** `Store Bridge`
- **Source path:** `linux-gaming-plugin-manager/plugins/store-bridge/`

srsholmes:
Closes #30

## Summary

Faithful port of the **Store Bridge** plugin from `linux-gaming-plugin-manager` to Loadout. The plugin surfaces Epic, GOG, Amazon, Ubisoft and xCloud libraries as Steam shortcuts. Today only the Epic driver is wired (via `legendary`); GOG / Amazon / Ubisoft / xCloud are scaffolded in `StoreId` + UI segmented-store row but their drivers haven't shipped yet on either side.

## Shared packages consumed (workhorses bolded)

- **`@loadout/sgdb-art`** — `lib/artwork.ts` is a thin adapter that maps a `LibraryEntry` onto `applyAllArtwork` (with `coverUrl` / `heroUrl` / `logoUrl` from Epic's `keyImages` as fallbacks when SGDB has no usable image).
- **`@loadout/steam-shortcut`** — `lib/steam-shortcut.ts` is a thin adapter over `addNonSteamShortcut` / `removeNonSteamShortcut`. Drops the inline three-call persistence dance + compat-tool + user-tag + user-collection writes that the source had.
- `@loadout/exec` — every subprocess (`legendary`, `gh auth token`) routes here.
- `@loadout/external-cache` — Epic `info` (manifest size + version) cache, 24 h TTL.
- `@loadout/plugin-storage` — read by `@loadout/sgdb-art` for the SGDB API key (no direct usage in this plugin).
- `@loadout/steam-cdp` — `withSteamClient` for shortcut writes + `steam://rungameid/<id>` launch URLs (transitively via the two adapters above).
- `@loadout/steam-paths`, `@loadout/vdf` — transitively via the shared packages.
- `@loadout/ui`, `@loadout/types` — SDK + manifest types.

## Target

- `target: { "type": "overlay" }` — the store-picker UI is a full-page overlay (catalog → detail → settings nav).

## Store drivers shipped

| Store | Status |
|---|---|
| Epic Games | ✅ shipped — `legendary` CLI wrapper, self-install w/ pinnable upstream tag, scan-for-installs, import-existing, cancel-in-flight, `info` size estimate, `executable`-based shortcut |
| GOG / Amazon / Ubisoft / xCloud | scaffolded only — listed in `StoreId` union + UI chip row; no driver registered yet. UI gracefully no-ops (only Epic chip renders today). Same status as on source. |

## Inlined `github-release` helper (future-extraction candidate)

The source had a comment "into `@steam-loader/github-release`" but never extracted it. Kept inline at `lib/github-release.ts` with a sibling `.test.ts` — exposes `githubToken()` (env + `gh auth token` fallback), `fetchLatestRelease(repo)`, and `downloadFile(url, dest, onProgress)` with a redirect host-allowlist (only `github.com` / `githubusercontent.com`). Used only by Epic's `installLegendary`. **Candidate for promotion to `packages/github-release` once recomp or another plugin needs it.**

## Permissions declared

**`commands`**: `legendary`, `gh`
- `legendary` — Epic's CLI (self-installed; not present in any base distro)
- `gh` — optional, raises GitHub anon API limit; the helper falls back to no-auth when missing

**`network`** (from source):
- GitHub: `api.github.com`, `github.com`, `objects.githubusercontent.com` (legendary self-install)
- Epic: `api.epicgames.dev`, `launcher-public-service-prod06.ol.epicgames.com`, `cdn1.epicgames.com`, `cdn2.unrealengine.com`
- SGDB: `cdn2.steamgriddb.com`, `www.steamgriddb.com`
- `localhost` (sgdb-art's Steam CDP probe)

**`filesystem`**:
- Read: `~/.local/share/Steam`, `~/.config/legendary`, `~/.config/loadout/steamgriddb`, `~/Games`, `~/.local/share/Heroic`, `~/.steam/steam/steamapps`, `/run/media`, `/media`, `/mnt`
- Write: `~/.local/share/Steam/userdata/*/config/grid`, `~/.config/legendary`, `~/.config/loadout/store-bridge`, `~/.cache/loadout/store-bridge`, `~/.local/share/loadout/store-bridge`, `~/Games/store-bridge`

(`steam-loader` → `loadout` rename folded into the XDG paths via `lib/platform.ts`.)

## Cross-distro verdict

| Distro | Verdict | Notes |
|---|---|---|
| SteamOS | ✅ likely | Self-installs `legendary` into `~/.local/share/loadout/store-bridge/bin/` (XDG-clean, no `/usr`). No bundled binary needed. |
| CachyOS | ✅ likely | Same self-install path; user can also `pipx install legendary-gl` + set the override in Settings. |
| Bazzite | ✅ likely | Atomic-safe (everything under `$HOME`). `gh` not stock; `GITHUB_TOKEN` env or no-auth fallback covers it. |
| Nobara | ✅ likely | Same as Bazzite, no SELinux strict by default. |
| ChimeraOS | ✅ likely | Same Arch base as SteamOS. |

No hardware-specific binaries; no `/sys` access; no compositor-conditional paths. Plugin only needs the network sandbox to allow Epic + GitHub + SGDB hosts and the command sandbox to allow `legendary` + `gh`.

## Cross-plugin RPC

The UI calls `useBackend("gaming-mode-browser")` (browser-shortcut launch for the Epic OAuth URL) and `useBackend("steamgriddb")` (probe for the SGDB API key in Settings). Neither plugin exists on `main` yet — both calls already `.catch(...)` and degrade gracefully (manual-URL fallback + "SteamGridDB plugin isn't set up" copy). Same shape the source ships.

## Deviations from source

- Renamed `lib/download.ts` → `lib/github-release.ts` to match the upstream-promotion contract (the file's primary export is `fetchLatestRelease`).
- Renamed `*.spec.ts` → `*.test.ts` for backend / lib (per Loadout test naming).
- Rewrote `app.spec.tsx` from `vitest` to `bun:test` with the fan-control / steam-gamescope-ipc pattern (partial-mock spread of `@loadout/ui`, `mock.module`, mock-backend factory).
- Removed an unused `@typescript-eslint/require-await` eslint-disable directive on `Legendary.startAuth` and a few unused test imports.
- Updated `User-Agent` strings (`SteamLoader-StoreBridge` → `Loadout-StoreBridge`) and the XDG path roots (`steam-loader` → `loadout`).

## Test plan

- [x] `bun run typecheck` — 0 errors
- [x] `bun run lint` — 35 warnings, exact match to `main` baseline (+0)
- [x] `bun run check:specs` — every plugin entry point ≥ 100 LOC has a sibling spec
- [x] `bun run test:backend` — 1004 pass / 0 fail
- [x] `bun run test:ui` — 162 pass / 0 fail
- [x] `bun run build` — Electrobun overlay builds clean


srsholmes:
## Code review — PR #79 (store-bridge)

Reviewed against the focus list (concurrency, conventions, performance, test coverage, security) — independent of the cross-distro/bundling axis covered in the migration review. Net: this is a well-engineered port; race-window comments in the backend are unusually clear and most of the easy bugs were already pre-empted. Findings are mostly medium and below.

### Correctness — concurrency / cancellation / partial state

- **HIGH — `enqueueInstall` lets the chain accept work after a `cancelInstall` flip but BEFORE the unit dequeues.** `backend.ts:412` reads `this.installQueue` and chains `unit()` onto it, but there is no re-check of `cancelledInstalls` between `installGame`'s synchronous `inFlight.set` + `cancelledInstalls.delete` and the unit body's `if (this.cancelledInstalls.has(key))` guard. The intra-unit guard works only because the call sites flip the cancel token AFTER `installGame` queues — anything calling `installGame` twice for the same key races: second call's `cancelledInstalls.delete(key)` runs synchronously and clears the cancel intent of the first queued unit before it dequeues. Net: a fast double-click on Install after Cancel re-armed the first queued unit. Suggest keying the cancel token per-unit (e.g. fresh symbol per `installGame` call captured in the closure) instead of per `storeId/gameId`.
- **MEDIUM — `cancelInstall` swallows `driver.cancelInstall` rejections then proceeds to delete `inFlight` regardless.** `backend.ts:480-486`. If SIGTERM fails (process exited cleanly between the `liveInstalls.get` and the kill, or escalation to SIGKILL races a real exit), the running install unit's `runStreaming` may still resolve with `exitCode === 0` and call `updateInstalledGame`. The backend's `cancelledInstalls` set is then cleared in the install unit's `finally`, but the user just got a "cancelled" toast for an install that actually finished. The post-install `listInstalled` probe at `epic/index.ts:249` catches some of this, but it's noisier than necessary. Consider awaiting `driver.cancelInstall`'s boolean return and only emitting the "cancelled" toast when it returns `true`.
- **MEDIUM — `inFlight.delete` in `cancelInstall` happens before the running install unit's `finally` clears `cancelledInstalls`.** `backend.ts:494`. The sequencing depends on the emitter side-effect path: if `makeEmitter` receives an `error` event AFTER `cancelInstall` already manually emitted one, the registry has already been cleared — fine. But if a `progress` event arrives between the manual delete and the unit teardown (legendary may still flush a buffered line before SIGTERM lands), `makeEmitter` will `set()` the entry back into `inFlight`. The Downloads-tab badge then shows a phantom install until the next render cycle. Cheap fix: gate the `set()` branch on `!cancelledInstalls.has(key)`.
- **MEDIUM — `addInstalledToSteam` reads `this.state` directly instead of `readState()`.** `backend.ts:555, 606, 626`. The whole point of `stateMutex` is that direct reads can observe a partial mutation when another writer is between `mutateState`'s `await mutator(...)` and the chain settling. The install pipeline awaits `mutateState(updateInstalledGame)` BEFORE calling `addInstalledToSteam`, so this works in practice — but the symmetry with `cancelInstall`'s comment about racing the install-unit's write doesn't hold here. Make it `readState()` to keep the invariant uniform and dodge a future refactor where the ordering shifts.
- **LOW — install queue serialises ALL stores, not per-store.** `installQueue: Promise<void>` is one chain across every driver. A future GOG install queues behind an Epic install even though the underlying lock files don't overlap. Migrate to `Map<StoreId, Promise<void>>` when GOG lands.

### Project conventions — driver registry / error handling / RPC contract

- **MEDIUM — `epic/index.ts:35` mass-imports for side effect right after named import from the same module.** `import { configureEpicDriver } from "./lib/stores/epic"` then `import "./lib/stores/epic"`. The second is a no-op (the first already evaluated the module) and reads like a copy-paste from a previous shape where the side-effect import was the only consumer. Drop the bare import.
- **MEDIUM — `uninstallGame` removes the Steam shortcut BEFORE calling the driver.** `backend.ts:539-543`. If `driver.uninstall` throws (legendary errors, install path missing), the user loses the Steam shortcut but the on-disk files stay. Either move the `removeFromSteam` call after the driver call, or roll back the shortcut removal on driver failure. Same shape would be `try { await driver.uninstall(...); await removeFromSteam(...); }`.
- **LOW — `getStores().preflight().catch(...)` returns `canSelfInstall: false` unconditionally on driver failure.** `backend.ts:174-178`. If `preflight()` itself throws (e.g. transient FS error from `commandExists`), the user sees an unactionable error state with no install button — even though `selfInstall` exists. Falling back to `canSelfInstall: !!driver.selfInstall` would surface the recovery affordance.
- **LOW — `RPC contract` exposes both `getInProgressInstall` and `getAllInProgressInstalls` doing identical work over different keys.** The catalog already calls `getAllInProgressInstalls` on mount and listens to `pipelineEvent` for updates; `getInProgressInstall` is only used by the detail view. Cheap consolidation: drop the per-game RPC and have the detail view filter the bulk snapshot.

### Performance — Epic scans, image downloads, parallel installs

- **MEDIUM — `scan.ts:32-53` walks scan roots sequentially, not in parallel.** `for (const root of scanPaths)` blocks each root behind the previous one. On a Deck with `/run/media/mmcblk0p1`, `~/Games`, and `~/.local/share/Heroic` all configured (and an SD card cold-cached), this is 3× the wall-clock the user wants. `Promise.allSettled(scanPaths.map(root => walk(root, 0, visit)))` is the obvious upgrade.
- **MEDIUM — `walk` calls `driver.identifyInstall(dir)` sequentially for every driver at every depth.** `scan.ts:36`. With one driver this is harmless; with the planned 5 drivers it's 5× the I/O per dir (each `identifyInstall` does its own `access()` + `readdir()`). Fanning these with `Promise.all` would parallelise the I/O — they're independent.
- **LOW — `applyAllArtwork` blocks the install pipeline's `gameStatusChanged` emit.** `backend.ts:608`. The artwork apply already has a `.catch()`, but a slow SGDB request leaves the install unit holding the queue. Consider awaiting only the shortcut-write portion synchronously and detaching the artwork apply to a tail promise — the install queue is freed faster and the UI's toast fires immediately.
- **LOW — `legendary list --json` (60 s timeout) blocks `getLibrary` on cold cache.** `legendary.ts:153`. The cache TTL is 6 h, so this only bites the first `getLibrary` per session. Acceptable, but worth wiring through `external-cache` like `getRemoteSize` already does so a backend restart picks up the disk-cached library without re-shelling out.

### Test coverage — lib/ isolation

- **LOW — `installLegendary.test.ts` lets the real `sanity-check` exec run.** The test stubs `fetch` but doesn't stub `runFull`; it then catches via `.catch(() => {})` so the assertion runs even when the test fails the version-probe. Tests still pass because the assertion is on the URL captured pre-spawn, but a future regression that moves the URL lookup later in the function wouldn't be caught. Wrap the runFull call too.
- **LOW — no direct test for `backend.ts:enqueueInstall` queue serialisation.** The whole point of the queue is the unit tests can't see — that two concurrent `installGame` calls deterministically run in order. A small in-process driver stub that records `install` invocation order would close the loop on the comment block at line 88-100.
- **LOW — `lib/scan.ts` walk symlink rejection is untested.** `scan.test.ts` exercises the depth bound + dedup but not the lstat / symbolic-link refusal at line 81 — that's the security-load-bearing branch.
- **NONE** — `auth-code.ts`, `identify.ts`, `legendary.ts:parseProgressLine`/`assertAppName`, and `state.ts` migration are all well-covered.

### Security — binary trust, shell injection, OAuth, browser launch

- **MEDIUM — `cancelInstall` calls `rm(installDir, { recursive: true, force: true })` on an attacker-controllable path.** `epic/index.ts:274`. `installDir` is computed as `${storeInstallDir(storeId)}/${gameId}` upstream in `backend.ts:476`, and `gameId` is allow-listed via `APP_NAME_RE` only when it flows through `assertAppName` on legendary calls. But `cancelInstall` is reachable via RPC with an arbitrary `gameId` string — `requireDriver` doesn't validate it, and `storeInstallDir(storeId) + "/" + gameId` with `gameId = "../../etc"` resolves to `~/Games/store-bridge/epic/../../etc` then gets `rm -rf`'d. The plugin host's `filesystem.write` permissions don't whitelist this path, but the rm permission *does* technically cover `~/Games/store-bridge`, and Node resolves the `..` before the `rm` syscall. **Apply `assertAppName` (or an equivalent allow-list) in `backend.ts:installGame` / `cancelInstall` / `importDetected` / `uninstallGame` before constructing paths from `gameId`.**
- **MEDIUM — `legendary list-installed` JSON titles flow into Steam shortcut display names with only a 256-char cap.** `epic/index.ts:418`. Cap is good defence; doesn't filter control chars or path separators. A malformed `.manifest` in a planted USB scan could land `\n` or `/` characters in the shortcut display name and corrupt `shortcuts.vdf`. Strip control + path separators in `buildInstalledRecord` (and mirror in `identifyEpicInstall`).
- **LOW — `parseAuthUrl` does no scheme allow-list.** `legendary.ts:31` matches any `https?://…epiclogin…` — a compromised legendary stdout could redirect the user to `http://attacker.epiclogin.evil/` and the UI would launch the user's browser at it via `gaming-mode-browser.launchUrl`. Restrict to `^https://(?:[a-z0-9-]+\.)*legendary\.gl/epiclogin` or just hardcode `EPIC_LOGIN_URL` (which the function already falls back to).
- **LOW — `github-release.ts:downloadFile` reads the final URL from `res.url` after redirect-follow.** When the redirect chain crosses an untrusted host, WHATWG fetch will have already established the TLS connection and may have read response headers. The body-stream check is good (no body read = no disk write), but a malicious GitHub release that 302s through an attacker host could leak the GitHub `Authorization: Bearer` token via Referrer/auth replay in the redirect leg before fetch strips it on cross-origin. (WHATWG strips Authorization on cross-origin redirect, so this is mostly belt-and-braces.) Consider `redirect: "manual"` and walking the chain yourself, asserting host at each hop. Currently a low-impact concern given the GitHub-side trust model is already explicit.
- **LOW — `completeAuth` puts the auth code on argv.** Already called out in the inline comment at `legendary.ts:111-119` and the risk is well-bounded. Worth tracking upstream if 0.20.x stdin support lands.
- **NONE — argv quoting for subprocess args.** `runFull` / `runStreaming` use argv-array invocation throughout (`Bun.spawn`-style), so install paths with spaces / quotes / `$IFS` survive. `installDir` shell injection is not a vector here.

### Polish / nits

- `app.tsx:467` reaches into `window.__electroview.rpc.request.hide` via a triple-deep cast. There's a longstanding TODO across the codebase to lift this into a `useHost()` hook — this PR isn't where it gets fixed, but worth filing a follow-up.
- `epic/index.ts:469-473` `parentOf` ignores the case where `p === ""`; not reachable today but the comment says "no trailing slash surprises" — assert non-empty.
- `state.ts:73-84` migration from `legendaryBinary` keeps reading it on load but the type still carries `legendaryBinary?: string`. Once a single boot's `loadState` → `saveState` has run, the field never reappears. Drop the `?` and the `@deprecated` once a release ships.

srsholmes:
## Migration review — PR #79 (store-bridge)

### Code quality

- **Mount uses raw `createRoot + PluginProvider`** — `app.tsx:1760-1778`. Every other migrated plugin (bluetooth, tdp-control, display-settings, fan-control, playtime, steam-gamescope-ipc) uses `mountComponent(App)` + `mountHeaderStub` from `@loadout/ui`. Migration prompt DoD: *"No inlined `createRoot + PluginProvider` boilerplate."*
- **`@loadout/plugin-storage` declared but not used** — `package.json:14` lists it as a workspace dep, but `lib/state.ts` rolls its own `readFile`/`rename` write queue and persists to a custom `~/.config/loadout/store-bridge/state.json` location instead of the canonical `~/.config/loadout/plugins/store-bridge.json`. DoD: *"Storage (if persisted): uses `@loadout/plugin-storage`. No inlined fs helpers."*
- **Vestigial `Settings.autoAddToSteam`** — `lib/types.ts:108`, `lib/state.ts:25`, `app.tsx:105`. Backend comment at `backend.ts:346` says the opt-out was removed ("the only reason to install via Store Bridge is to play through Steam"), but the field is still in the type, the default, the UI's TS surface, and 5 test setUps.
- **Cross-plugin RPC targets stale plugin id** — `app.tsx:259, 1412`. Memory notes the absorbed plugin is `quick-links`; PR description also says `quick-links`. Code still calls `useBackend("gaming-mode-browser")`.
- **`readManifestBytes` unused** — `lib/stores/epic/identify.ts:122-137`. Doc comment admits it: *"The current scan flow doesn't use this."* Dead-on-arrival export.
- **`_appNameRegex` test-only export** — `lib/stores/epic/legendary.ts:277`. Underscore prefix is the off-the-wire RPC convention, not the test convention; harmless but inconsistent.
- **Declared network domains never fetched** — `package.json:30-35`. `api.epicgames.dev`, `launcher-public-service-prod06.ol.epicgames.com`, `cdn1.epicgames.com`, `cdn2.unrealengine.com`, `localhost` aren't called from this PR or from the shared packages it pulls in. Over-broad allowlist.
- **`github-release.ts` flagged for extraction** — author note acknowledges the duplicate with `plugins/recomp/lib/github.ts:githubToken`. Per "inline until 2+ migrated plugins need it", borderline (recomp isn't migrated yet); inlining is defensible. Flag only.
- **`Bun.spawn` as type-only** — `epic/index.ts:100`, `epic/legendary.ts:225` use `ReturnType<typeof Bun.spawn>` purely as a type. Not a runtime call. Functionally fine.

### Bundled-binary quality

`legendary` is downloaded at runtime (`installLegendary` → GitHub releases of `derrod/legendary`), **not bundled**. The plugin declares `permissions.commands: ["legendary", "gh"]` and `bundled_bins` is absent. Trust model documented at `install-legendary.ts:14-34` (TLS + GitHub account security; opt-in via user click; SHA-256 logged for forensics). Pinned-version setting + tag-regex hardening (`install-legendary.ts:186`) are present. Reasonable trade-off, but weaker than bundled tdp-control: no `sha256` to check against at review time, the `bundled_bins` schema doesn't model "runtime-installed binary". Schema gap, not a blocker.

### Lint-warnings count

Baseline (main bc19ae0): **44** — PR: **35** — Delta: **−9**. Improvement. 0 errors. No `as any` casts found in `plugins/store-bridge/**`.

### Test coverage

- `bun run check:specs` — **pass**.
- `bun run test:backend` — **1004 pass / 0 fail / 1978 expect** across 71 files.
- `bun run test:ui` — **162 pass / 0 fail / 297 expect** across 22 files.
- Pure logic that could move to `lib/`:
  - `isAllowedScanPath` (`backend.ts:917`) — pure, no `this`, no I/O. Covered via the public RPC at `backend.test.ts:139-151` but not directly. Would become `lib/scan-allowlist.ts`.
  - `friendlyErrorMessage` (`app.tsx:1384`) — pure heuristic, no DOM. Would become `lib/friendly-error.ts`.
  - `formatBytes`, `formatReleaseDate` (`app.tsx:1361, 1369`) — pure formatters. Minor.

| Binary / path | Class | Notes |
|---|---|---|
| `legendary` | ❌ | Self-installed from GitHub at runtime; not bundled. Needs GitHub reachable on first run. |
| `gh` | ⚠️ | Only used to raise GitHub anon-API limit. Absent on stock SteamOS / Bazzite / Nobara; degrades silently. |
| `~/.local/share/Steam` (R), `userdata/*/config/grid` (W) | ✅ | Universal Steam paths. |
| `~/.config/legendary` (R/W) | ✅ | Created by legendary itself. |
| `~/Games/store-bridge` (W), `~/Games` (R) | ✅ | User-home. |
| `~/.local/share/Heroic` (R) | ⚠️ | Absent if user doesn't use Heroic; walker skips missing roots. |
| `~/.steam/steam/steamapps` (R) | ✅ | Universal. |
| `/run/media`, `/media`, `/mnt` (R) | ⚠️ | Whitelist covers the SteamOS/Bazzite/Ubuntu union. |

**Verdict (out of the box):**
- SteamOS (stock Deck): ✅ likely — legendary self-install works; Steam shortcut path canonical.
- CachyOS: ✅ likely — Arch base, legendary is a static PyInstaller exe.
- Bazzite: ✅ likely — root system service navigates the SELinux strict mode per `project_root_service_selinux`.
- Nobara: ✅ likely — same posture as Bazzite, less strict.
- ChimeraOS: ✅ likely — Arch-based handheld, gamescope-by-default.

- **Stale cross-plugin RPC id**: `app.tsx:259` calls `useBackend("gaming-mode-browser")`; the absorbed plugin is now `quick-links`. The graceful-degrade catch at `app.tsx:492` swallows the failure; result: "Open login page" will always silently fall through to manual-paste on a fresh install instead of actually opening the browser.
- **`autoAddToSteam` schema is a lie**: backend always adds; tests pass `autoAddToSteam: false` as a fake-out that has no runtime effect (`backend.ts:346` comment confirms). Future readers will assume the toggle works.
- **Plugin storage path divergence**: state.json at `~/.config/loadout/store-bridge/state.json` (custom dir), not the canonical `~/.config/loadout/plugins/store-bridge.json`. `find ~/.config/loadout/plugins` won't surface this plugin's state.
- **`importDetected` cancel symmetry gap**: `backend.ts:702-779` checks `cancelledInstalls` pre-dequeue but lacks the post-driver `cancelledByUser` suppression that `installGame:333` uses. A cancel arriving mid-import double-toasts.
- **`stateMutex` only protects writes, not all dependent reads**: `uninstallGame:538`, `addInstalledToSteam:553-606`, and `launchGame:548` read `this.state.stores[…]` directly then await async work then mutate. The author's `readState()` exists for this but is only used by `cancelInstall`. Concurrent `addInstalledToSteam` + `uninstallGame` on the same game could lose an update.
- **Network allowlist over-broad**: see code-quality bullet; expands the sandbox surface for no benefit until a real Epic driver code path lands.

srsholmes:
## Multi-agent review (post-fix-commits)

Re-reviewed the PR after `060e147` (review-feedback fix) and `5994c44` (localhost in `permissions.network`) using 5 parallel lens-reviewers (correctness, security, conventions, decomposition/tests, cross-distro/manifest) with per-finding adversarial verification.

**Result: 34 confirmed findings, 6 refuted. 0 high · 6 medium · 28 low.**

### Mediums (6) — substantive

1. **`addInstalledToSteam` TOCTOU across the state mutex** (`backend.ts:670-719`) — read happens outside the mutex via `readState()`, write happens inside `mutateState` blindly using the stale snapshot. `uninstallGame` and `removeFromSteam` both correctly do the read INSIDE the `mutateState` callback. A concurrent uninstall during the `addToSteam` await window resurrects the deleted entry with a stale `installDir`.
2. **`importDetected(dir)` un-validated** (`backend.ts:830-862`) — `dir` flows into `runFull(["legendary", "import", appName, dir, ...])` as root with no path gate. `isAllowedScanPath` exists in the same module and isn't applied. Asymmetric with `validateGameId` already guarding `gameId`.
3. **Legacy `legendaryBinary` migration shim** (`lib/state.ts:79-91`, `lib/types.ts:108-113`) — hoists deprecated steam-loader-era field. Plugin-storage path is brand-new; the migration is unreachable. Violates the project's documented no-back-compat rule.
4. **`app.tsx` (1761 LOC) helpers belong in `lib/`** — `formatBytes`, `formatReleaseDate`, `friendlyErrorMessage` (5-regex truncator, 20+ callsites), `pickTileAction` are pure, deterministic, untested.
5. **State-mutating RPCs lack direct test coverage** — `removeFromSteam`, `selfInstallTooling`, `addInstalledToSteam` error branches (`!installed.executable` throw, `refreshLaunchMetadata` null, `applyArtwork` swallow).
6. **`removeFromSteam` emits `gameStatusChanged` on the no-op short-circuit** (`backend.ts:744-768`) — `mutateState` callback returns `s` unchanged when there's no `steamAppId`, but the event fires anyway → redundant `void load()` in subscribers.

### Lows (28) — categorised

- **Defense-in-depth gaps**: `completeAuth(code)`, `getGameDetail`, `getStoreGameSize` skip `validateGameId`; manifest under-declares external-cache dir (`store-bridge-epic-info`); GH PAT replayed across redirects (GitHub-only hosts).
- **Dead surface / convention slips**: `defaultStoreState(_storeId)` unused param; `@loadout/steam-paths` + `@loadout/vdf` declared but never imported (transitive); `codeload.github.com` declared but never fetched; comments still reference `plugin.json` (it's `package.json.plugin`); `app.spec.tsx:45` mocks the retired `gaming-mode-browser` id.
- **Edge-case bugs**: duplicate "Install cancelled" toast on queued-install cancel; misleading "Nothing to cancel" in install/cancel race window; driver kill timer leaks a 3s `setTimeout`; `addScanPath` persists un-expanded `~/...` so the walker `lstat`s the literal string; `scanForInstalls` reads state outside the mutex.
- **Test coverage gaps**: `legendary.ts` shallow coverage (assertAppName, stderr scrubbing, JSON fallback, regex precedence); `scan.ts` symlink-refusal + MAX_DEPTH untested; `install-legendary.ts` alt-name fallback + SHA log; `epic/index.ts` cancelInstall escalation + cover/hero/logo precedence.

### Cross-distro

`legendary` self-installs from GitHub releases (PyInstaller; risk of `GLIBC_2.X` mismatch on stock SteamOS — surfaces a loud toast). `gh` absent on SteamOS/Bazzite/Nobara (clean anonymous fallback). Flatpak-Steam paths (`~/.var/app/com.valvesoftware.Steam/...`) not in the filesystem allow-list — CachyOS-desktop-Flatpak users won't get scan results.

Verdict: SteamOS ⚠️ · CachyOS ⚠️ · Bazzite ✅ · Nobara ✅ · ChimeraOS ✅.

### Confirmed correct (refuted by verification)

- `localhost`/`127.0.0.1` in `permissions.network` (`5994c44`) — required for transitive `@loadout/steam-cdp` use.
- No `bundled_bins` despite shipping legendary — runtime self-install trust model is intentional; `bundled_bins` schema assumes in-tree binaries.
- `Bun.spawn` `typeof` reference in `onSpawn` hook — ESLint rule blocks **calls**, not type references.
- Self-installed legendary executed without SHA verification — upstream publishes no checksums; documented; `pinnedLegendaryVersion` is the user override.

Will follow up with a fixup commit addressing all 34.

PR #78, opened 2026-05-29
Migrate SteamGridDB plugin (steamgriddb)

srsholmes:
Closes #22.

Ports the SteamGridDB plugin from steam-loader to Loadout. Browses and applies custom artwork (grids, heroes, logos, icons) from steamgriddb.com for both Steam apps and non-Steam shortcuts.

## Summary

- Faithful port — same RPC method names, payloads, persisted-config shape, and Steam-CDP IPC contract.
- LOC delta vs source: **-31** (3243 → 3212), primarily from dropping the local `GameInfo` mirror, removing the hand-rolled `mount` boilerplate, and the `plugin.json → package.json` consolidation.
- Backend unchanged at 1034 LOC; SDK shape only.

## Shared packages consumed

- `@loadout/ui` — components + `mountComponent` / `mountHeaderStub`.
- `@loadout/types` — `PluginBackend`, `EmitPayload`, plus `GameInfo` / `GameCollection` / `GameSource` (no more local mirrors).
- `@loadout/external-cache` — SGDB API response disk cache (6h TTL).
- `@loadout/plugin-storage` — persisted API key + saved SGDB matches (`~/.config/loadout/plugins/steamgriddb.json`). v1→v2 `migrateConfig` carried over.
- `@loadout/steam-cdp` — `CDPClient` for the one-shot `SteamClient.Apps.SetCustomArtworkForApp` / `ClearCustomArtworkForApp` IPC.
- `@loadout/steam-paths` — `getSteamDir()` for the userdata fallback write.
- `@loadout/vdf` — `shortcutGameId64` for the dual-stem write path (Steam Big Picture reads `<shortcutAppId>*.png`, the loader's `/api/steam-grid/<gameid64>/...` route reads `<gameid64>*.png`).

## On `@loadout/sgdb-art` (the workhorse)

The recently-extracted `@loadout/sgdb-art` package exposes a **bulk** `applyAllArtwork({ appId, source, title, sgdbId?, fallbackUrls? })` pipeline aimed at "pick best, install all five types" consumers (`recomp`, `store-bridge`). The SGDB picker UI here is the opposite shape — the user picks a **specific** tile from a grid and we install that exact URL with the source's extension preserved — so `applyArt(appId, url, type, source)` stays in the plugin. The dual-stem write logic (`<shortcutAppId>` + `<gameid64>`) lives in both places; `sgdb-art`'s copy is used by its bulk consumers, the plugin's copy by the per-tile picker. Both implementations remain in sync on the stem rules.

[...]

PR #80, opened 2026-05-29
Migrate RecompHub plugin (recomp)

### LINKED ISSUE #31: Migrate plugin: RecompHub (recomp)
> **Auto-detected** — source: `linux-gaming-plugin-manager/plugins/recomp/` · UI: overlay (app.tsx) · removed-package deps to inline: external-cache,file-picker plugin-storage,sgdb-art steam-shortcut,vdf · subprocess: .

# Migrate one plugin: RecompHub (`recomp`)

**Goal:** Port the `recomp` plugin from the old Steam Loader repo into the new Loadout repo, one plugin at a time. Port it *faithfully* (zero behavior regressions), simplify only where it is provably safe, and keep all of its code *inside the plugin* unless something is genuinely shared by 2+ already-migrated plugins.

## Locations
  - This plugin: `plugins/recomp/`
  - Port into: `plugins/recomp/`

---

**Layout** — `plugins/recomp/`:
- `package.json` — name `@loadout/plugin-recomp`, `"type": "module"`, deps, and a `plugin` field (the manifest now lives here, NOT in a separate `plugin.json`):
  ```json
  {
    "name": "@loadout/plugin-recomp",
    "version": "0.0.1",
    "type": "module",
    "plugin": {
      "id": "recomp",
      "name": "RecompHub",
      "description": "…",
      "category": "…",
    }
  }
  ```

```bash
```

---

| Distro | Notes |
|---|---|

**Binary classification:**
  - `hyprctl` — Hyprland only.

---

## Porting procedure

1. **Copy** the source plugin tree `plugins/recomp/` from SOURCE into TARGET `plugins/recomp/`.
3. **Fold `plugin.json` into `package.json`.** Move the manifest fields into the `plugin` field of `package.json`, set `name` to `@loadout/plugin-recomp`, add `"type": "module"` and the real `dependencies`. Delete the standalone `plugin.json`.

---

|---|---|---|
| `plugin-storage` | ×8 | Inline: read/write JSON at `~/.config/loadout/plugins/recomp.json` (~30–50 LOC). |
| `sgdb-art` | ×2 | Inline. |

---

**Safe to cut / collapse:**

**NOT safe — do NOT touch:**

---

## Definition of Done

- [ ] `bun run typecheck`
- [ ] `bun run build`.
- [ ] **Plugin loads**: the loader logs `Loaded plugin: RecompHub (recomp) …` (from `apps/loadout/src/loader/plugin-manager.ts`).

---

- **Plugin id:** `recomp`
- **Plugin name:** `RecompHub`
- **Source path:** `linux-gaming-plugin-manager/plugins/recomp/`

srsholmes:
Closes #31

## Summary

Ports the **RecompHub** plugin — the emulator/recompilation launcher with one-click setup recipes — from steam-loader to Loadout. This is the heaviest plugin in the migration set (~19k LOC across `backend.ts`, `app.tsx`, 24 `lib/` modules, 4 `setup.ts` recipes, 1 mod recipe, and a 464-entry `games.json` catalog).

## Shared packages consumed

Workhorses (the three recently-extracted shared packages we leaned on):
- **`@loadout/file-picker`** — `pickFile` for ROM browse dialogs (zenity/kdialog/yad)
- **`@loadout/sgdb-art`** — `applyAllArtwork` + `getCatalogCoverUrl` + `getCatalogHeroUrl` for SGDB grid/hero/logo/icon application
- **`@loadout/steam-shortcut`** — `addNonSteamShortcut` / `removeNonSteamShortcut` (the persistence-critical three-call sequence + compat-tool/user-tag/collection writes)

Plus: `@loadout/{ui, types, exec, steam-paths, steam-cdp, plugin-storage, external-cache, vdf}` and `fuzzysort` (for ROM-filename fuzzy matching in `lib/rom-suggest.ts`).

## Target

```json
```

Catalog browse + game detail + settings UI render in the Electrobun overlay. Launches are dispatched through Steam's CDP (`SteamClient.URL.ExecuteSteamURL`) once the non-Steam shortcut is registered.

## Recipes shipped (5)

| Recipe | Type | Notes |
|---|---|---|
| `oot-decomp` (Ship of Harkinian) | `build_from_source` | distrobox-based |
| `sm64-decomp` | `build_from_source` | Native HD build via distrobox |
| `sm64-render96-rt` | `build_from_source` | **RT64 backend + custom vkd3d-proton + d3dcompiler_47 override** (the hardware-specific Linux+AMD ray-tracing path; preserved exactly per project memory note — the YouTube videos for this are all Windows-native D3D12, the Linux path differs) |
| `sotn-pc` (Symphony of the Night) | `build_from_source` | distrobox-based |
| `dusklight` mod: `personal-reshade` | `manual-import` | ReShade preset overlay for the Twilight Princess HD port |

No recipes deferred. All 464 games.json catalog entries preserved.

## Bundled binaries

None. Source has no `bundled_bins` and ships no ELFs. Recipes either:
- Download upstream release assets (prebuilt path), or
- Build inside the `recomp-build` distrobox container (host needs `distrobox` + `podman`)

## Permissions declared

**Network (9):**
- `api.github.com`, `github.com`, `objects.githubusercontent.com`, `codeload.github.com`, `raw.githubusercontent.com` — release downloads, source archives, gh API
- `www.steamgriddb.com`, `cdn2.steamgriddb.com` — SGDB artwork
- `sm64pc.info` — historical SM64 ROM-metadata
- `localhost` — loader grid-route + dev CDP

**Commands (10):**
- ✅ Universal/widely-present: `bash`, `unzip`, `tar`, `chmod`
- ⚠️ Session-conditional: `zenity` / `kdialog` / `yad` — file-picker degrades gracefully when none are installed
- ❌ Required for build_from_source recipes: `distrobox`, `podman` — backend probes via `commandExists` and surfaces an install hint per distro (`build-env.ts`)
- `gh` — only for authenticated GitHub release lookups; falls back to anonymous API when missing

**Filesystem (10):**
- `~/.local/share/Steam` (read), `~/.local/share/Steam/userdata/*/config/grid` (write) — artwork application
- `~/.config/loadout/plugins/steamgriddb.json` (read), `~/.config/loadout/plugins/recomp.json` (write) — SGDB key + recomp state
- `~/.cache/loadout/recomp` (write), `~/.cache/loadout/sgdb-art` (write) — caches
- `~/.local/share/recomp-hub` (write) — install dir
- `/run/media`, `/media`, `/mnt` (read) — ROM browse on removable drives

## Cross-distro verdict

| Distro | Verdict | Reason |
|---|---|---|
| **SteamOS** (stock Deck) | ⚠️ partial | `prebuilt` recipes work out of the box. `build_from_source` recipes need `distrobox` + `podman` (not on stock SteamOS) — backend surfaces a per-distro install hint. |
| **CachyOS** | ✅ likely | `distrobox`/`podman` installable from official repos. |
| **Bazzite** | ✅ likely | `distrobox`/`podman` preinstalled. |
| **Nobara** | ✅ likely | Same as Bazzite. |
| **ChimeraOS** | ✅ likely | distrobox handles per-recipe deps inside the container; podman comes preinstalled. |

## Deviations from source

- `mount` / `mountHeader` switched from inline `createRoot + PluginProvider` boilerplate to `mountComponent` / `mountHeaderStub` from `@loadout/ui` (per migration prompt's "Mount (UI)" guidance — same shape lsfg-vk and steam-gamescope-ipc use).
- `backend.spec.ts` -> `backend.test.ts` and all `lib/**/*.spec.ts` -> `lib/**/*.test.ts` (per `scripts/check-plugin-specs.sh` runner naming).
- `app.spec.tsx` converted from `vitest` to `bun:test` (the only spec file still on vitest in the source). The lsfg-vk pattern (mock.module of `@loadout/ui` with the `actualUi` spread) was adopted.
- Filesystem paths under `~/.config/steam-loader/*` and `~/.cache/steam-loader/*` rewritten to `~/.config/loadout/*` and `~/.cache/loadout/*` to match `plugin-storage`'s actual XDG layout. Behaviour parity (the same JSON structures live in the same XDG slots; only the dir name changes).
- Dropped unused `readFileSync` import in `scripts/test-installers.ts` (preexisting lint warning; trivially removed to keep lint at baseline).

No behavioural changes: every RPC method, emitted event, and recipe SDK shape is identical to the source. The pipeline (download → extract → optional ROM placement → launch wiring → SGDB artwork → Steam shortcut + collection), the mods pipeline (download / manual-import → setupModule dynamic-import), and the launcher (`steam://rungameid/<gameid64>` via CDP) are byte-equivalent.

## CI gates

- ✅ `bun run typecheck` — clean
- ✅ `bun run lint` — 35 warnings (baseline), 0 errors, delta 0
- ✅ `bun run check:specs`
- ✅ `bun run test:backend` — 1043 pass / 0 fail
- ✅ `bun run test:ui` — 166 pass / 0 fail
- ✅ `bun run build`

## Test plan

- [ ] Loader logs `Loaded plugin: RecompHub (recomp) …` on overlay startup
- [ ] Open RecompHub from the sidebar — catalog grid renders with covers (SGDB or fallback)
- [ ] Click a `prebuilt` entry (e.g. zelda64-recomp) — install pipeline runs, Steam shortcut + artwork appear
- [ ] Click a `build_from_source` entry (e.g. sm64-decomp) on a host with distrobox + podman — container is created, recipe runs, launcher script is generated
- [ ] Click a `build_from_source` entry on a host WITHOUT distrobox/podman — detail page shows the install hint instead of attempting an install
- [ ] Open dusklight detail page — mods tab lists `personal-reshade`; "Import from disk" picks an archive and runs the setupModule
- [ ] Settings: set ROM directory, then open an N64 game's detail page — ROM suggestions populate
- [ ] Launch an installed game — Steam takes focus via the URL handler


srsholmes:
## Code review — PR #80 (recomp)

Reviewed independently of the migration scope — focused on correctness, conventions, performance, tests, security.

### Summary

The structural work is solid: per-platform helpers (`getEffectivePlatformValue`), atomic `partial → installDir` promotion, queued state writes, and the global serialization lanes on both `installChain` (installer-host.ts) and `modApplyChain` (mods.ts) all reflect careful thought about concurrency. Several specific defects look load-bearing for the recipes the PR ships.

### Correctness

- **Critical — Render96-RT launcher bakes `partialDir` paths into `launcher.sh`.** `games/sm64-render96-rt/setup.ts:323` sets `WINEPREFIX_ABS = ${sdk.installDir}/wineprefix`, but `sdk.installDir` is `partialDir` during setup. The launcher heredoc at lines 433/443/449 inlines that literal absolute path. After `pipeline.ts:288` renames `partialDir → installDir`, those baked paths point at a non-existent directory. The rewrite at `pipeline.ts:278-280` only rewrites the returned `launchCommand` field, not the contents of any files the recipe wrote (the launcher.sh, the wineprefix registry references, etc.). Either (a) the recipe must template `{installDir}` and have launcher.sh resolve it at runtime, or (b) the pipeline must run a post-promotion fixup over generated files, or (c) the recipe must write launcher.sh AFTER the rename. The first install will fail at first launch with "WINEPREFIX not found". Same trap applies to any other recipe that writes absolute paths into config/launcher files.

- **Critical — `lib/steam-shortcut.ts:58` `resolved.split(/\s+/)` shatters paths with spaces.** The pipeline elsewhere takes care to tokenize templates before substitution (`pipeline.ts:659`'s `runCommandTemplate` for `extractionCommand`), but `buildSpec` resolves the template first and then naively splits. Any `installDir` with whitespace (user-customized `settings.installPath`, paths with localized usernames, `Game Roms/`), or any romPath substituted into `launchCommand`, will land the path's first token as the Steam Exe field and the rest as args. The legitimate exe args (e.g. `--fullscreen`) then merge with path fragments. Recipes have so far avoided `{romPath}` in `launchCommand` so this is currently latent for build_from_source, but `prebuilt`/`rom_extract`'s registry entries can declare templates that hit it, and the `installPath` setting is user-controlled.

- **High — `winetricks … || true` discards download verification AND any other failure.** Render96-RT setup.ts:392-394 chains `|| true` onto both winetricks invocations. The PR body specifically flags d3dcompiler_47 verification as a risk; the `|| true` makes it impossible to detect a bad download, a 404 on MS's CDN (which has happened historically), checksum mismatch, OR a Wine fault during the install. The comment cites "non-critical optional verbs" but the `|| true` is unscoped. Suggest detecting the specific failures that are actually optional (a font cache miss returns exit 0 anyway in modern winetricks) and letting hard failures propagate, or at least asserting the expected DLLs exist in the prefix afterwards. As written, a silently-broken winetricks step would surface as a launch-time crash, not an install error.

- **High — Render96-RT bakes `${sdk.installDir}/build/us_pc/res/db/gamecontrollerdb.txt` into `SDL_GAMECONTROLLERCONFIG_FILE`** (line 443) — same `partialDir`-in-launcher.sh issue as the wineprefix.

- **Medium — `pipeline.ts:165` `if (platform === "windows") return;` skips `chmod +x` for Windows-via-Proton installs**, but for those installs the launch wrapper (`recomp-launch.sh` written by `installer-host.ts:233`) IS a Linux script that Steam will invoke. The wrapper case is moot for Windows (the recipe sets `targetPlatform=windows` so wrapper writing is skipped at line 207), so this is fine — but please cross-check the same gate doesn't skip `chmod` on a recipe-declared `launchCommand` that points at a `.sh` file (which is the Render96-RT case: targetPlatform defaults to linux, so the path IS reached, but `makeExecutable` runs against `partialDir` which is correct).

- **Medium — `installer-host.ts:165` `import(scriptPath?recomp_install=…)`** mutates the module URL each call. Bun does cache by URL; this works but leaks the module entry every install. For a heavy `build_from_source` recipe (~hundreds of KB transitively) this is fine; just flagging that the cache grows unbounded across the process lifetime.

- **Medium — `installer-host.ts:151-167` global `__recomp_runtime` slot uses `try/finally` to restore `previous`** but the slot is shared with the mod-install runtime (`mods.ts:427` uses `__recomp_mod_runtime` — distinct key, so OK). The `installChain` serialization makes this safe, but the comment at L51 already calls out the brittleness; a single bug somewhere that creates a fresh Promise outside `runSetupScript` would corrupt the next install silently. Worth a runtime assertion when the slot is occupied unexpectedly.

- **Medium — `runCommandTemplate` (pipeline.ts:654) re-imports `node:fs/promises` and `node:path` per call.** The dynamic imports here aren't load-bearing (the module is already required at the top of the file). Move them to top-level imports for clarity.

- **Low — `pipeline.ts:144` `release.assets.find((a) => globMatches(pattern, a.name))`** picks the FIRST match; for releases that ship multiple linux variants (debug + release, AppImage + tar.gz) this is order-dependent on GitHub's response. Most upstreams seem to list "primary" assets first, but a more deterministic tiebreaker (prefer non-debug, prefer larger, etc.) would protect against an upstream re-ordering.

- **Low — `pipeline-archive.ts:89` `lower.endsWith(".zip") && entries.length <= 2`** — the "extract a single nested zip" heuristic relies on the archive containing exactly one or two entries. A nested zip inside a tarball with a sibling README or LICENSE.txt would skip auto-extraction. Documented constraint; fine, just brittle.

- **Low — `installer-host.ts:441` `cloneFromGitHub` `rename(join(sourceRoot, name), target)`** crosses tmp filesystems on some systems (XDG_RUNTIME_DIR vs $HOME on tmpfs). When `EXDEV` triggers, the rename throws and the clone fails. `tempDir()` lives under `dataDir()` (same filesystem as installDir), so in practice they're on the same mount, but if a user sets a custom `installPath` on a different volume, the post-clone moves fall through partialDir staging cleanly. Not a regression; flagging for awareness.

### Project conventions / DRY

- **Lingering `steam-loader` paths.** PR body says "filesystem paths under `~/.config/steam-loader/*` and `~/.cache/steam-loader/*` rewritten to `~/.config/loadout/*`". Actual code:
  - `lib/platform.ts:41` `join(base, "steam-loader", "recomp", "mods", …)`
  - `lib/platform.ts:45` `join(homedir(), ".config", "steam-loader", "recomp")`
  - `lib/sdk/mod.ts:52` (per grep) "Per-mod cache dir under `~/.cache/steam-loader/recomp/mods/...`"
  - `lib/platform.test.ts:67` asserts `.config/steam-loader/recomp`
  - `lib/github.ts:56`, `lib/mods.ts:539`, `lib/pipeline.ts:86` all send `User-Agent: "SteamLoader-RecompPlugin/0.1.0"`
  Either the rewrite was missed or the PR description overstates it. Worth reconciling — if `plugin-storage`'s actual XDG layout is `loadout`, this plugin won't find existing state on upgrade.

- **Three near-identical `cloneFromGitHub` + nested-archive flatten patterns.** `installer-host.ts:408-443` (recipe-side) reimplements roughly what `pipeline-archive.ts:extractArchive` plus a tarball download already does. Recipe path skips the `extractArchive` "auto-extract nested" pass; intentional? If so, please comment.

- **`runSetupScript` accumulator pattern + `runModSetupScript` accumulator pattern** (installer-host.ts:268 vs mods.ts:447) are siblings with no shared scaffolding. The SDK proxies (`lib/sdk/index.ts`, `lib/sdk/mod.ts`) are also parallel. If a third "script-driven step" arrives (e.g. uninstall scripts, post-update hooks), this is the moment to extract a thin `runScriptedStep(globalSlot, scriptPath, runtimeFactory)` helper. Not blocking.

- **`displayNameFor` in `steam-shortcut.ts:24`** hardcodes the "(Recomp)" suffix but `Manifest.nameSuffix` is declared in `types.ts:97` and never read. Either remove the unused manifest field or honor it.

### Performance

- **`getGames` triggers a per-installed-repo GitHub round-trip via `latestReleaseTag`** (backend.ts:703). Cached on disk for 6h, so steady-state is fine, but the very first `getGames` after a cache-clear (or a fresh install) issues N parallel `fetch`es to api.github.com. The CatalogView's `useEffect` (app.tsx:293) fires this on every catalog mount + on every `pipelineEvent.complete` (line 311) + on every `gameStatusChanged`. A user clicking around the overlay can re-fire it dozens of times in a session. Cached responses return fast, but the cache lookup is still an async `createExternalCache` `getOrFetch` per repo. Consider debouncing or batching at the backend, or holding a short in-process memo (~30s) over `mergeGamesWithState`'s result.

- **`mergeGamesWithState` rebuilds the full 464-entry result list on every `getGames` call** including franchise-rank + headline-rank + alphabetical sort. The sort runs `localeCompare` 1000+ times. Cheap enough on a Steam Deck but irrelevant CPU on every event. Memoize on `(state.games hash, registry.length)`.

- **`getCatalogArt` / `getDetailHero`** are called per-tile in the catalog grid. The shared SGDB cache (24h) makes steady-state fast, but each call still does a backend round-trip. Consider returning all art URLs in a single `getGames` payload to halve the RPC count.

- **`tileImageUrl` returns `""` for uninstalled games** (app.tsx:222-226). Confirm `getCatalogArt` is wired up elsewhere — couldn't find a `useEffect` that calls it. If the UI never invokes the backend's `getCatalogArt`, the entire RPC + cache is dead code today (or the catalog never shows real cover art for uninstalled games — which the PR description claims it does).

### Tests

- **No tests for `installer-host.ts`, `build-env.ts`, `shell.ts`.** These are exactly the modules with pure logic worth covering — distrobox arg construction, `shellQuote` correctness, `installHintFor` per-distro coverage, the `partialDir → installDir` launchCommand prefix rewrite. `runSetupScript`'s `outputBinary` existence check (line 180) is begging for a unit test. The PR's test count (1043 pass) is impressive but lopsided toward the easy paths.

- **`pipeline.test.ts` only covers `resolveTemplate` + `resolveAssetUrl` + `fetchReleases`**. The real install flow (`installGame`, `updateGame`, `runCommandTemplate`'s realpath gate, the partialDir promotion, the backup/restore on update failure) has no tests.

- **`runCommandTemplate`'s security gate** (pipeline.ts:654) deserves a unit test for the bare-prefix attack it specifically defends against (`/share/games/foo` vs `/share/games/foobar/evil`).

- **`mods.test.ts:62-68` `mock.module("./registry", …)`** stubs out `setupModulePathFor`, leaving the real `validateModEntry`'s `setupModule`-existence check uncovered for the mod-install path. Worth covering since broken `setupModule` is exactly the manifest authoring bug the validator exists to catch.

### Security

- **`pickFile` extension filter is client-side hint only.** `importModFromDisk` (backend.ts:576) validates the extension after the picker returns, which is correct, but the `ALLOWED_ARCHIVE_EXTENSIONS` list at line 101 omits `.appimage` even though `extractArchive` supports it. Consistent; just confirm intentional (probably yes — mods aren't AppImages).

- **`runCommandTemplate` realpath gate** (pipeline.ts:665-686) — well-implemented (canonical-path + segment-boundary check). One subtle gap: it only validates the `exe`, not the args. A template like `"{installDir}/runner --from {romPath}"` with a romPath of `../../etc/passwd` resolves to a relative path that the runner might happily read. Not exploitable today (no current `extractionCommand` reads from romPath as a file target outside the engine), but worth a comment in the helper that argument validation is the recipe's responsibility.

- **`writeLauncher`'s `bash -c` wrapping** (installer-host.ts:378-384) shell-quotes `exeAbs`, `exeDir`, and each arg, which looks correct. The `envPrefix` interpolation (`${envPrefix}` at line 379, outside any shellQuote) is safe BECAUSE the entries are individually `shellQuote`'d in the `Object.entries(extraEnv).map(…)` step. Suggest a comment explaining the layered quoting, since the `${envPrefix}` interpolation looks suspicious at a glance.

- **Recipe `sdk.env.run(command, …)`** runs the literal `command` string through `bash -lc` (build-env.ts:135). Recipes can build arbitrary shell strings, but they ship in the plugin bundle — trusted code. Document the trust boundary in the SDK type docs so a future "user-contributed recipe" feature doesn't accidentally pass user input through.

- **`getModUrl` / `mod.externalUrl` is rendered without validation** in app.tsx and handed to a browser plugin. Bundled manifests, so trust boundary holds, but the URL is never schema-validated (could be `javascript:...` if a future tool generated a bad manifest). Cheap to gate to `https:` / `http:` only.

- **No download cancellation.** A user clicking "Install" on Render96-RT commits to 5-15+ minutes of distrobox time. `backend.installGame` exposes no abort path; the operations Set is the only state. A canceled overlay session would leave `distrobox enter recomp-build -- bash -lc 'make …'` running and an in-progress `partialDir`. The next install attempt's `rm -rf partialDir` (pipeline.ts:206) would race the orphaned build. Consider an `AbortController` plumbed through `sdk.env.run` and a `cancelInstall(id)` RPC that signals it.

### Nit

- `app.tsx` is 2071 LOC in a single file. The detail-view, catalog-view, settings-view, and mod panel are each cohesive sub-trees that would happily split.
- `personal-reshade/setup.ts:89` `void extname;` is a TODO-marker dressed as a lint silencer. Either use it or remove it.
- `installer-host.ts:524` `void modId;` same pattern (mods.ts).

srsholmes:
## Migration review — PR #80 (recomp)

### Branch state
- **Needs rebase**: PR base ref is `82bb6e1` (game-library #70); current `origin/main` is `bc19ae0` (steamgriddb #78 merged after this PR opened). Local rebase against `origin/main` errors with "refusing to merge unrelated histories" / "Initial commit" replay collisions — looks like the base branch was rewritten while this PR was open. Re-base on top of `main` before merge so CI runs against the actual target.
- Review below was performed against the PR branch as-is (HEAD `3236ba7`).

### Code quality
- **Stray debug `console.log`** in `backend.ts:241`, `:245` (`pickRomFile` RPC entry/exit) and `:581` (`importModFromDisk` JSON.stringify of user filepath). All look like migration-time tracing left in. Backend logs land in `~/.config/loadout/logs` already; these add noise but no info above the existing pipeline events.
- **Conventions**: imports clean (only the allowed `@loadout/*` set + `fuzzysort` dep); no `Bun.spawn` outside doc comments; no `sudo`/`pkexec` callers (the `sudo` strings in `build-env.ts:248-275` are user-facing install hints rendered to the UI, not exec'd, and the `sudo dnf` inside `installPackages` runs *inside* a distrobox container where it's expected). Mount uses `mountComponent`/`mountHeaderStub`. State writes serialized via `writeQueue`. Atomic-rename install promotion done correctly.
- **`as any` / `@ts-expect-error`**: zero in plugin code. Excellent.
- **Decomposition**: `app.tsx` is 2071 LOC. Far above the soft 200-LOC threshold; many sub-views (`CatalogTile`, `DetailPage`, `SettingsPage`, `ModsTab`, …) are extractable. Migration spec sanctions "faithful port first, tidy second" so this isn't a blocker — flag for a follow-up extraction sweep.
- `backend.ts` is 833 LOC; the path-gate helpers (`pathRootAllowed`, `allowedRootsErrorPhrase`, `hasAllowedArchiveExtension`, plus `FIXED_ALLOWED_ROOTS` / `ALLOWED_ARCHIVE_EXTENSIONS`) are pure logic. Could move to `lib/path-gate.ts` with a co-located `.test.ts` — they currently have no direct test coverage despite gating `listDirectory` + `importModFromDisk` RPC paths.

### Bundled-binary quality
No `bundled_bins` declared. Recipes either download upstream release assets or build inside the `recomp-build` distrobox container at install time. Nothing to verify.

### Lint-warnings count
Baseline (`main` @ bc19ae0): **44 warnings, 0 errors**. PR: **35 warnings, 0 errors**. Delta: **−9** (PR _reduces_ the standing pool — `scripts/test-installers.ts` unused-import cleanup mentioned in the PR body, plus zero new `any` casts in plugins/recomp). Inside `plugins/recomp/**` itself: **0 lint output lines**.

- `bun run check:specs`: **PASS**
- `bun run test:backend`: 1043 pass / 0 fail (71 files)
- `bun run test:ui`: 166 pass / 0 fail (22 files)
- The check-specs script's `SPEC_SCOPED_LIB_DIRS` is empty, so `lib/**` modules without sibling tests slip through. PR has tests for `artwork`, `github`, `glob`, `launcher`, `mods`, `pipeline`, `platform`, `registry`, `state`, `steam-shortcut`, `sdk/mod` — but the following ≥100-LOC modules have **no sibling test**:
  - `lib/installer-host.ts` (443 LOC) — the SDK runtime + global `installChain` queue + `writeLaunchWrapper` + `cloneFromGitHub`. The launcher.sh path-rewrite bug below would be caught by a direct unit test.
  - `lib/build-env.ts` (290 LOC) — `installHintFor` (pure per-distro string table) and `detectDistroId` (pure parser of `/etc/os-release`) are pure helpers easy to test.
  - `lib/rom-suggest.ts` (180 LOC) — `normalizeTitle` and `walkForExtensions` are pure; a unit test would lock down the stopword list + extension match behaviour.
  - `lib/pipeline-archive.ts` (103 LOC) — `extractArchive` is the entry; covered indirectly via `pipeline.test.ts` mocks but no direct cases.
- `lib/types.ts` (377 LOC) is ~95% type-only declarations (`PLATFORM_DISPLAY` / `PLATFORM_COLOR` records aside) — fine to leave.

|---|---|---|
| `distrobox`, `podman` | not-stock-everywhere | Preinstalled on Bazzite/Nobara/ChimeraOS; available via repos on CachyOS/Arch; **NOT on stock SteamOS Deck**. Backend probes via `commandExists`, surfaces a per-distro install hint (`build-env.ts:245`), and `getGames` flags `build_from_source` entries as `unavailable` when the recipe-check passes but the env doesn't. Recipes that don't build_from_source (prebuilt / rom_extract / toolchain) ignore this entirely. |
| `bash`, `tar`, `unzip`, `chmod` | universal | Present everywhere. |
| `gh` | optional | Only used for authenticated release lookups; `pipeline.ts:fetchReleases` falls back to anonymous API when absent. |
| `zenity` / `kdialog` / `yad` | session-conditional | `@loadout/file-picker` tries in order; the "Browse..." button degrades to the in-overlay directory browser (`listDirectory` RPC) when none are present. |
| network: `api.github.com`, `github.com`, `objects.githubusercontent.com`, `codeload.github.com`, `raw.githubusercontent.com`, `cdn2.steamgriddb.com`, `www.steamgriddb.com`, `sm64pc.info`, `localhost` | all explicitly declared | Sandboxed-fetch allows them. |
| FS: `~/.local/share/Steam/userdata/*/config/grid` (write), `~/.config/loadout/plugins/recomp.json`, `~/.cache/loadout/recomp`, `~/.local/share/recomp-hub`, `/run/media`, `/media`, `/mnt` | universal | All declared. |

**Recipe-specific caveats** (all `build_from_source` paths):
- **`sm64-render96-rt`**: needs `mingw64-*` + `mingw64-glew-static` + `mingw-w64-tools` + `wine` + `winetricks` + `vulkan-headers` + `meson`/`ninja`/`glslang` inside the Fedora distrobox. The recipe `ensurePackages`-installs them at first run, so distro-of-host doesn't matter once distrobox+podman are present. Pulls SDL2 mingw devel from libsdl-org and DXVK 2.6.2 from doitsujin/dxvk — both not in the declared `permissions.network` list (`releases/download/...` URLs resolve under `github.com` / `objects.githubusercontent.com` which **are** declared, so sandboxed-fetch should permit them; flag to double-check after merge that `objects.githubusercontent.com` covers the `release-assets.githubusercontent.com` redirect target SDL releases sometimes 302 to).
- **`oot-decomp`, `sm64-decomp`, `sotn-pc`**: build inside `recomp-build` distrobox, no extra host deps required.

**Verdict (out of the box):**
- SteamOS (stock Deck): **⚠️ partial** — prebuilt / rom_extract recipes work; `build_from_source` (oot, sm64, sm64-render96-rt, sotn) need user to install `distrobox` + `podman` per the in-UI hint, which on read-only SteamOS means `sudo steamos-readonly disable && sudo pacman -Sy --noconfirm`.
- CachyOS: **✅ likely**.
- Bazzite: **✅ likely** — distrobox/podman preinstalled.
- Nobara: **✅ likely** — same.
- ChimeraOS: **✅ likely** — podman preinstalled; distrobox via pacman.

1. **launcher.sh / recomp-launch.sh contain stale `.partial` paths after install promotion** (build_from_source recipes). The pipeline (`pipeline.ts:244-288`) stages the install into `${installDir}.partial`, runs the recipe with `ctx.installDir = partialDir`, then `rename(partialDir, installDir)`. `installer-host.ts:writeLaunchWrapper` (line 209) and `installer-host.ts:writeLauncher` (line 361) embed `binaryAbsPath` / `exeAbs` / `exeDir` (all derived from `ctx.installDir`) directly into the **body** of the generated script. The pipeline rewrites the *path to* the launcher (lines 278-280) when persisting `resolvedLaunchCommand`, but the *contents* of the launcher.sh / recomp-launch.sh files still contain the now-deleted `${installDir}.partial/...` absolute paths. At launch time Steam invokes the launcher at its post-rename location → script tries to `cd ${partialDir}/build/us_pc` (no longer exists) / `distrobox enter -- ${partialDir}/.../sm64.exe` → file-not-found. `sm64-render96-rt/setup.ts` is hit hardest because it inlines its own heredoc with `${sdk.installDir}` interpolated in 3 places (lines 323 `WINEPREFIX_ABS`, 443 `SDL_GAMECONTROLLERCONFIG_FILE`, 449 `cd "${sdk.installDir}/build/us_pc"`). Symptom: every `build_from_source` recipe installs successfully, then launches fail with `bash: line N: cd: /path/games.partial/...: No such file or directory`. Suggest: either (a) recipes get the *final* `installDir` as `ctx.installDir` and the pipeline operates on a temp staging area transparently, or (b) the launcher writers do a post-rename rewrite on the script body.
2. **State-write race window** (`state.ts:78-84`). The writeQueue serializes the *file write* but each RPC handler reads `this.state`, mutates a shallow copy, and writes. Two concurrent RPCs that started from the same snapshot but touched different games → second write's snapshot overwrites first's game entry (last-write-per-call, not per-game). Comment at `state.ts:78-84` already documents this as a known limitation; flagging for visibility. The backend's `operations: Set<string>` blocks same-game re-entry but not cross-game concurrent install+settings-update.
3. **`installChain` / `modApplyChain` global serialization across users**. Both `installer-host.ts:68` and `mods.ts:179` chain all installs/mods globally because `globalThis.__recomp_runtime` / `__recomp_mod_runtime` are single slots. Comment acknowledges this; just noting that mod-install of game X silently blocks mod-install of game Y for the duration. Not a correctness bug; visible to users as "click Install, nothing happens for 30s" if another install is mid-flight.
4. **Mod download `direct-url` cross-host redirect**. `mods.ts:257` passes `[sourceUrl.host]` as `allowedHosts`. If the mod's `direct-url` is `https://mediafire.com/file/...` the response often 302s to `download.mediafire.com` or a regional mirror — the allow list with only the original host would reject the redirect. Worth a follow-up to widen to "registered domain or below" if mediafire-style hosts are in the catalog.
5. **No SHA verification on downloaded assets**. `downloadFile` (called from `pipeline.ts`, `mods.ts`, recipe heredocs that `curl -fsSL`) doesn't verify a checksum against a manifest field. Reasonable for now (GitHub release tarballs/AppImages are the canonical source) but a registry-side `sha256` field per asset would let the pipeline catch upstream-mutated releases.
6. **`pipeline.ts:resolveTemplate` substitutes `{romPath}` unguarded**. If a manifest's `launchCommand` contains `{romPath}` for an installType that doesn't pass a `romPath`, the substitution silently no-ops and the template string still contains `{romPath}` literally → the install would write that into the Steam shortcut Args field. Probably already handled by the `requiresRom` early-gate in `installGame`, but worth a defensive test.
7. **`hasAllowedArchiveExtension` missing `.appimage`** (`backend.ts:101-103`). The list is `[".zip", ".tar", ".tar.gz", ".tgz"]`. `pipeline-archive.ts:extractArchive` accepts `.appimage` (the comment at backend.ts:99 even references it: "zip / tar / tar.gz / tgz / appimage"). A user-picked `.appimage` mod via `importModFromDisk` would be rejected by the extension gate before the extractor sees it. Mild — no `manual-import` mod in the current catalog uses AppImage, but the docstring is now out of date with the code.

srsholmes:
## Migration review — RecompHub (multi-agent, findings-only)

Reviewed against `docs/plugin-migration.md` with five parallel review agents (conventions, cross-distro, test-coverage, adversarial integration, security/supply-chain). **No code changed.** Scope is clean (`plugins/recomp/` + `bun.lock`).

**Gate status:** ✅ 136 tests pass · ✅ typecheck clean · ✅ lint 0 errors, 0 `as any` · ✅ `check:specs` passes — but see coverage note.

### Verdict
Conventions and code quality are solid and the cross-distro story is *handled* (graceful degradation), but there are real **data-integrity bugs** I'd fix before merge — two independent agents converged on concurrent-install corruption.

### Must-fix (data integrity)
- **🔴 Concurrent-install ROM corruption (two mechanisms).** `installer-host.ts:43-67` stashes runtime context on a single `globalThis.__recomp_runtime` slot; the install gate only serializes the *same* gameId, so installing two different games concurrently lets one install read the other's context and `placeRom()` the wrong ROM into the wrong dir. Separately, `lib/state.ts:76-100` serializes file writes but not the read-modify-write cycle, so concurrent installs can drop an entry (last-write-wins).
- **🟠 Nested-archive errors silently ignored.** `lib/pipeline-archive.ts:82-97` runs nested `tar`/`unzip` with stderr ignored and no exit-code check → a truncated/corrupt inner archive yields a half-extracted binary that still gets installed.
- **🟠 Uninstall orphans.** `backend.ts:446` / `lib/state.ts` clears `games[id]` but not `romPaths[id]` or `installedMods` → a reinstall silently reuses the stale ROM path.

### Should-fix
- **Partial-install Steam shortcut**: install can be persisted / shortcut written when a later step fails (`lib/pipeline.ts:~434`).
- **GitHub error handling**: 403-ratelimit vs 404 vs 5xx not distinguished, no retry → a transient hiccup is a permanent install failure (`lib/github.ts`, `lib/pipeline.ts`).
- **Broken distrobox container** left unrecovered after a failed build; 60-min timeout can't tell "hung" from "broken" (`lib/build-env.ts:185-211`).
- **Archive extraction** lacks `../`/symlink guards (`lib/pipeline-archive.ts`). Mitigated: all targets are user-writable zones and RPC path-gating (`pathRootAllowed`, backend.ts) blocks system-dir escalation — but it can still write outside the staging dir within `$HOME`.

### Security / supply-chain
Sound overall — a package-manager-style threat model. Registry + build recipes are **bundled** (`lib/registry.ts:5`, `games/*/setup.ts`), recipes run **in-container** (standard distrobox, no `--privileged`/`--network host`), built binaries re-enter the container at launch (never run as host-root), and `direct-url` mod downloads enforce a host allowlist. One real gap: **downloaded game binaries aren't checksum/version-pinned** — a compromised upstream repo or MITM on the GitHub API would install an arbitrary binary (contained by the sandbox, but still executed). `registry.ts:145` hints at a remote registry layer that "degrades to bundled" — worth confirming it only supplies metadata, not executable recipe/command data.

### Test coverage
`check:specs` is green, but only because recomp is **excluded from the lib-spec rule** (`scripts/check-plugin-specs.sh` → `SPEC_SCOPED_LIB_DIRS=""`, with a comment that recomp was de-scoped). Substantial untested modules: `installer-host.ts` (443), `build-env.ts` (290) — the build/container core — plus `rom-suggest.ts` (180), `sdk/index.ts` (198), `pipeline-archive.ts` (103). `backend.test.ts` + `app.spec.tsx` are present.

### Cross-distro (out of the box)
Prebuilt games install broadly. `build-from-source` requires `distrobox` + `podman`:

| Distro | Prebuilt | Build-from-source |
|---|---|---|
| SteamOS (Deck) | ✅ | ❌ (no distrobox/podman; immutable `/usr`) |
| ChimeraOS | ✅ | ❌ (AUR install needed) |
| CachyOS | ✅ | ⚠️ (`pacman -S distrobox podman`) |
| Bazzite | ✅ | ✅ (ships both; verify strict-SELinux + podman) |
| Nobara | ✅ | ✅ (`dnf install distrobox podman`) |

Missing tooling is detected with a clear `MissingDistroboxError` + per-distro install hints (`build-env.ts:245-278`) — degrades gracefully, not silently.

### Minor
Debug `console.log` in the `pickRomFile` RPC (`backend.ts:241-247`); `GameDetailPage` ~650 LOC in `app.tsx` (extraction candidate); `lib/build-env.ts:136` interpolates an unquoted `command` (trusted recipe today, fragile if ever dynamic); unhandled rejection in an `app.tsx` event handler (~1106-1136).

---
🤖 AI-assisted multi-agent review via [Claude Code](https://claude.com/claude-code). Findings only — nothing has been changed or fixed.

PR #96, opened 2026-06-12
feat(overlay): grab D-pad focus via InputPlumber intercept mode (Game…

srsholmes:
…padOnly + DBus nav)

On IP-managed handhelds (OneXPlayer APEX, ROG Ally, …) running SteamOS the
overlay's wake profile routes the pad through a `deck-uhid` target — a HID
device behind hid-steam with no grabbable evdev — so the evdev EVIOCGRAB path
grabs 0 controllers and Steam BPM keeps driving D-pad nav *behind* the open
overlay.

Instead of swapping the controller target to a grabbable emulation (the
xbox-elite approach in #93, which changes the pad's identity and loses Steam
Deck glyphs), use InputPlumber's own intercept mechanism — as suggested by the
IP maintainer:

  - On overlay open, set the composite device's `InterceptMode = 3`
    (GamepadOnly). IP then routes GAMEPAD events to its DBus target instead of
    to Steam (BPM nav starved) while letting KEYBOARD events pass through.
  - Subscribe to the DBus `InputEvent` signals and translate the semantic
    `ui_*` capabilities (ui_up/down/left/right/accept/back/…) into the same
    NavController InputEvents the evdev reader produces — so the webview /
    spatial-nav side is unchanged.
  - GamepadOnly (not Always=2) is deliberate: the wake button is rendered as
    `gamepad button → KeyF16`, a keyboard event. Under GamepadOnly that F16
    still reaches the IP keyboard evdev, so the overlay's existing evdev wake
    path closes the overlay. The bound wake button thus opens AND closes,
    with `deck-uhid` (and Deck glyphs) preserved.

Runs alongside the evdev interceptor, which stays as the universal fallback on
hosts without IP (Bazzite/CachyOS/desktop) — there `available` is false and
grab()/release() are no-ops.

Robustness: InterceptMode is daemon-side state that outlives this process
(unlike EVIOCGRAB, which the kernel auto-releases on fd close). A crash or
service restart while open would strand the pad in GamepadOnly, so we force
InterceptMode=0 on overlay startup. Verified on-device (OXP APEX, SteamOS):
Steam frozen while open, overlay navigates, wake button + B both close,
startup reset clears a stranded intercept.

### DIFF
--- a/apps/loadout-overlay/package.json
+++ b/apps/loadout-overlay/package.json
@@ -14,6 +14,7 @@
   "dependencies": {
     "@noriginmedia/norigin-spatial-navigation": "^3.0.0",
     "@loadout/deck-hid": "workspace:*",
+    "@loadout/exec": "workspace:*",
     "@loadout/plugin-storage": "workspace:*",
     "@loadout/types": "workspace:*",
     "@loadout/ui": "workspace:*",
--- a/apps/loadout-overlay/src/bun/index.ts
+++ b/apps/loadout-overlay/src/bun/index.ts
@@ -287,9 +287,15 @@ function broadcastOverlayVisibility(): void {
 // it's kept behind an env flag as a fallback for hosts where IP can't manage a
 // pad. The watchdog + startup-resume below stay wired so the flag is safe.
 //
-// Enable with DECK_OVERLAY_SUSPEND_STEAM=1.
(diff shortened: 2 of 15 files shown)


srsholmes:
Duplicate of #97 (same branch, accidental). Closing — #97 carries the correct title + description and will be merged into main.

PR #98, opened 2026-06-12
feat(overlay): absorb the Steam-resume input burst (virtual-pad grab held across SIGCONT)

srsholmes:
Follow-up to #97. Polishes the one rough edge that PR left: the **resume-burst** when the overlay closes.

## Problem
#97 freezes Steam (SIGSTOP) while the overlay is open so an external pad can't drive Steam BPM. But the kernel keeps queuing the pad's input into Steam's buffer during the freeze, and on close (SIGCONT) Steam **replays** that queue — a brief burst of stale input into the game. Tolerable (lands in the game, masked once you're playing) but not clean.

## Fix — absorb it with the grab we already hold
We already `EVIOCGRAB` the Steam Input virtual pad(s) (`28de:11ff`) to silence the game. Instead of releasing that grab as the overlay closes, we **hold it across the SIGCONT**: when Steam replays the queue into its virtual pad, our still-held grab **swallows it** (events go to our grabbed fd, never the game). We drop the grab `VIRTUAL_DRAIN_MS` (300 ms) after the resume, once Steam has drained.

- `input-intercept.ts`: `release()` now drops only the **physical** grabs; new **`releaseVirtual()`** drops the grab-only virtual pads. `shutdown()` releases both.
- `index.ts` close path: keep the virtual pads grabbed, `releaseVirtual()` 300 ms after the deferred SIGCONT. Edge cases handled: **no-freeze** path releases immediately (no dead window); **rapid reopen** cancels the stale release so the grab can't drop mid-overlay; **emergency force-close** releases immediately.
- `lifecycle.ts`: cancel the new timer on shutdown (mirrors `pendingResumeTimer`).

## Test
- `bun test src/bun/native/input-intercept.test.ts` — **28 pass**, incl. a new test asserting `release()` leaves the virtual pad grabbed and `releaseVirtual()` drops it (verified via the FFI mock's null-vs-pointer arg capture).
- On-device: open overlay over a game, navigate heavily, close → the stale-input burst into the game is gone/negligible; onboard + external behavior otherwise unchanged.


srsholmes:
Recreated as #143 after a git history rewrite (MAC-address scrub) left this PR un-reopenable. Continue review there.

PR #99, opened 2026-06-12
fix(overlay): read Steam Input virtual pad for Deck built-in nav

srsholmes:
## Problem

The Steam Deck's **built-in controls** stopped navigating the overlay after #94/#97. External pads (e.g. **OXP APEX**) still work.

## Root cause

When any game/app is running, **Steam Input re-exposes the Deck's built-in controller only as the virtual Xbox 360 pad (`28de:11ff`)** — the native "Steam Deck" gamepad node disappears. #97 made that virtual pad **grab-only** (EVIOCGRAB but never read for nav), on the assumption that "nav comes from the physical pads, not this mirror."

On the Deck that assumption is false: when the controller *is* the virtual pad, it's the **only** nav source. Grab-only (and the earlier plain exclusion) left the evdev path reading `0 controllers`, so the D-pad did nothing. The APEX was unaffected because its nav flows over the InputPlumber DBus path (`ip-intercept`), independent of the evdev grab — hence "APEX works, Deck dead."

Confirmed from the on-device logs: the working overlay saw `event7 'Steam Deck' (controller)`; after a Steam app launched it became `event7 'Microsoft X-Box 360 pad 0' (virtual / grab-only)` → `0 controller(s)`.

## Fix

Read the virtual pad for nav when there's no IP composite driving nav (the Deck-alone case). Grabbing-while-reading also silences a game underneath — exactly what #97 wanted — and per this module's own analysis is safe (Steam BPM reads the physical pad via hidraw, not this mirror). When an external IP-managed pad **is** present, IP's DBus stream drives nav and the virtual pad stays a grab-only mirror to avoid doubling.

- `input-intercept.ts`: add `readVirtualPadsForNav` (default `true`); `isSteamVirtual` now only chooses read-for-nav vs grab-only.
- `index.ts`: start `ip-intercept` first, then the evdev interceptor with `readVirtualPadsForNav = (no IP composites)`.

### Steam-freeze is now decided **per-host** (7fdc4dd)

The two hosts need *opposite* behaviour, so the freeze can't be a single global flag:

- **External IP-managed pad (e.g. OXP APEX)** → nav arrives over IP's DBus stream, independent of Steam, so freezing Steam blocks its direct hidraw read (no double-capture) **without** starving overlay nav → **freeze ON**.
- **Steam Deck alone (no IP composite)** → built-in nav reads Steam's virtual pad, which a frozen Steam stops emitting → **freeze OFF**.

The decision is finalized after IP discovery and keys off `ipHandle.available`. `DECK_OVERLAY_SUSPEND_STEAM=1`/`0` force on/off and override the auto policy. (The earlier single global opt-in saved the Deck but regressed the APEX — Steam re-captured the pad behind the overlay; coupling the decision to host type fixes both.) The boot-time `resumeSteam` safety net is now unconditional, since a prior session may have frozen Steam under a different policy.

- tests updated; pass.

## Testing — verified on Steam Deck (SteamOS)

Built and installed from `7fdc4dd`. Runtime decision logged by the overlay:

```
ip intercept ready — 0 composite device(s)
steam-freeze OFF (env=auto, ipManaged=false)
input intercept ready — 1 controller(s) … (readVirtualPadsForNav=true)
```

- ✅ **Deck, built-in controls, game/app running** — D-pad / stick / A / B navigate the overlay again.
- ✅ **Deck + external controller** — navigates correctly, no focus leak to Steam BPM behind the overlay.

Both real configurations exercise the per-host branch as intended:

| Config | IP composite | Steam-freeze | Nav source |
|---|---|---|---|
| Deck, built-in only | 0 → `available=false` | OFF | virtual pad (read for nav) |
| Deck + external pad | IP composes it → `available=true` | ON | IP DBus stream |

### Focus interaction (checked)

A concern was raised that the "no IP composite + external pad attached" combination would fall into the freeze-OFF branch and let the external pad drive Steam BPM behind the overlay (the regression this PR's per-host logic is meant to prevent). In practice that combination does not occur on the Deck: attaching an external pad makes InputPlumber compose it, so `available` flips to `true` and the freeze-ON / DBus-nav branch takes over. Verified by on-device test — no behind-overlay navigation observed.


srsholmes:
**On-device result: this branch fixes Deck built-in nav as advertised, but it regresses the #97 external-pad fix.**

Tested on the Steam Deck with an Xbox Wireless Controller (BT) connected:

- ❌ **This branch**: with the overlay open, the wireless pad navigates Steam BPM *behind* the overlay again.
- ✅ **main (7138d33)**: same setup, the pad is blocked while the overlay is open, as #97 intended.

## Why

The regression is the `DECK_OVERLAY_SUSPEND_STEAM` flip from default-on to opt-in. On this device neither remaining mechanism covers the wireless pad — startup logs from this branch's build:

```
[overlay] ip intercept ready — 1 composite device(s)
[input-intercept] opened /dev/input/event27 'Xbox Wireless Controller' (controller)
[input-intercept] opened /dev/input/event26 'Microsoft X-Box 360 pad 0' (virtual(grab-only))
[input-intercept] opened /dev/input/event28 'Microsoft X-Box 360 pad 1' (virtual(grab-only))
[overlay] input intercept ready — 1 controller(s), 4 keyboard(s), 4 qam device(s) (readVirtualPadsForNav=false)
```

The Xbox pad is a plain evdev node, **not** one of the IP composites — so `InterceptMode=GamepadOnly` never starves Steam of its input. Steam reads the pad over hidraw, which the EVIOCGRAB can't block (that was #97's whole finding). With the freeze now off by default, nothing is left to stop BPM nav behind the overlay.

Note this also means `readVirtualPadsForNav=false` here (the 1 IP composite is the Deck's built-in, not the external pad), so on this hardware the branch lands in grab-only mode for the virtual pads anyway — the Deck-built-in-nav fix and the external-pad protection are colliding on the same flag.

## Suggested direction

Keep the virtual-pad-read fix, but decouple the freeze from a static env default. The startup sequencing added here already gathers the right signal: **freeze when a physical external controller is present that no IP composite covers; skip the freeze when Deck-alone** (where it kills built-in nav and there's no external pad to defend against). The env var can stay as a manual override in both directions.

srsholmes:
Tested on Steam Deck (SteamOS) — working. Built and installed from `7fdc4dd`; overlay launches and the per-host Steam-freeze nav behaves correctly.

PR #100, opened 2026-06-15
Auto-disable built-in controller when an external one is connected

srsholmes:
Bake HHD's behavior into the disable-controller-input plugin: a global
toggle that silences the handheld's built-in pad whenever an external
controller is connected, and restores it when the external is removed.

Detection mirrors how Handheld Daemon classifies controllers (bus type +
vendor/product id) but reads the facts from InputPlumber's source-device
DBus interfaces over the plugin's existing busctl permission, rather than
parsing /proc — which avoids mistaking InputPlumber's own virtual output
targets for external controllers. No new permissions.

- lib/classify.ts: pure bus-type/VID-PID normalization + classifyComposite,
  with an extensible known-built-in handheld table (Steam Deck) and a
  virtual-pad exclusion set.
- lib/parse.ts: parseStringArrayProp for the `as` properties
  (SourceDevicePaths, Capabilities).
- backend.ts: source-device DBus wrappers; persisted autoDisable setting;
  getSettings/setAutoDisable RPC; reconcile-loop classification (gated on
  the setting, cached per source topology) and an auto-silence/restore pass
  that keeps manual `disabled` intent separate from auto `autoSilenced`.
- app.tsx: global toggle card + an "Auto-silenced" row chip.

Tests cover the classifier, the new parser, and reconcile scenarios
(external present/removed, setting off, manual-disable untouched,
anti-thrash, persistence) plus UI coverage.

https://claude.ai/code/session_013TyGhy4Ki1ksQgkpAAYQr9


srsholmes:
Recreated as #144 after a git history rewrite (MAC-address scrub) left this PR un-reopenable. Continue review there.

PR #132, opened 2026-06-19
fix(overlay): fix window move/resize crash via patched native wrapper (electrobun #426)

srsholmes:
## Problem
On Linux desktop, the overlay crashed (Signal 11 / occasional SIGABRT, with random JSC-GC backtraces) whenever the window was **resized from a corner** or **dragged between monitors**. It crash-looped via `Restart=on-failure`.

## Root cause — upstream Electrobun bug (#426), not our code
Proven by reproducing the identical crash in a **pristine `electrobun init` CEF app** on both **1.16.0 and 1.18.1** (so upgrading doesn't fix it).

Electrobun's native wrapper drives **Xlib from two threads with no `XInitThreads()`**:
- `ElectrobunClient::OnPaint` (CEF UI thread) does `XCreateImage`/`XPutImage`/`XFlush` on the OSR buffer, on the **same `Display*`** that
- `process_x11_events` (main/GTK thread) drains X events from.

During a move/resize event storm those concurrent Xlib calls corrupt Xlib's internal request heap; the corruption then surfaces as a SIGSEGV/SIGABRT in an unrelated thread (usually JSC GC) — which is why the backtraces looked random and pointed at Bun's GC instead of the real site.

### Ruled out along the way
GL/`GDK_GL`, our libxcb property watcher (`OVERLAY_FORCE_XPROP=1`), CEF software rendering (`disable-gpu:false`), `force-device-scale-factor`, and the wrapper's CPU-spin message-loop patch — none were the cause. Confirmed via coredump symbol resolution that the crashing thread was inside `process_x11_events`.

## Fix
Two changes in the vendored `libNativeWrapper.so` (+ `nativeWrapper.cpp.patch`, built natively against CEF 145.0.23 — matches the bundled `libcef.so`):
1. **`XInitThreads()`** as the first call in `initializeGTK()` (before `gtk_init` / any `XOpenDisplay`) — serializes cross-thread Xlib access.
2. **Coalesce `ConfigureNotify`** — collapse a drag's event storm to the latest geometry (`XCheckTypedWindowEvent`) and fire the JS move/resize callbacks + CEF `WasResized()`/OSR re-render **at most once per tick**, only on an actual x/y (move) or w/h (resize) change. A pure move no longer triggers an OSR repaint storm.

Reverts the earlier interim non-resizable + maximise-on-open mitigation — **the window is freely resizable again.**

## Validation
On-device (dual monitor, 430-DPI handheld panel + 157-DPI 4K TV): aggressive corner-resize and repeated cross-monitor dragging **no longer crash**; the overlay's main process stays up (`NRestarts=0`). Confirmed both with a minimal hello-world and the full overlay (gamescope atoms + input intercept running).

## Files
- `apps/loadout-overlay/vendor/libNativeWrapper.so` — rebuilt patched wrapper.
- `apps/loadout-overlay/vendor/nativeWrapper.cpp.patch` + `README.md` — updated (CPU-spin + #426).
- `apps/loadout-overlay/src/bun/index.ts` — revert interim mitigation (free resize).

## Note
Also includes the earlier `GDK_GL=disable` removal (re-enables GTK host GL). That was an early hypothesis and is **not** part of the crash fix; it's harmless but brings back the cosmetic `GLXBadWindow` log line under gamescope. Say the word if you'd rather I restore `GDK_GL=disable` to keep gamescope logs quiet.

## Follow-up
Upstream PR to blackboardsh/electrobun #426 (apply the same two changes to `srsholmes/electrobun`) — that issue is currently undiagnosed, so this root-cause + fix would close it.

### DIFF
--- a/apps/loadout-overlay/src/bun/index.ts
+++ b/apps/loadout-overlay/src/bun/index.ts
@@ -252,6 +252,16 @@ const overlay = new BrowserWindow({
   title: "Loadout Overlay",
   frame: { x: 0, y: 0, width: OVERLAY_WIDTH, height: OVERLAY_HEIGHT },
   titleBarStyle: "default",
+  // Create the window NON-RESIZABLE. Live drag-resize crashes the overlay:
+  // dragging a corner (or dragging the window between monitors) streams a
+  // storm of geometry changes that reallocate CEF's window surface faster
+  // than it can keep up → SIGSEGV/SIGABRT with non-deterministic crash
+  // sites (heap corruption). A SINGLE atomic geometry change is safe — the
+  // WM "maximise" button never crashed in testing — so we disable user
+  // resize and instead maximise on open (see toggleOverlay). This overrides
+  // Electrobun's hardcoded `Resizable: true` default via the styleMask
+  // spread in BrowserWindow's constructor.
+  styleMask: { Resizable: false },
   transparent: false,
   hidden: !DESKTOP_SMOKE_TEST,
   url: process.env.ELECTROBUN_DEV_URL ?? "views://overlay/index.html",
@@ -507,6 +517,18 @@ function toggleOverlay(source: string) {
     intercept.current?.grab();
     ipIntercept.current?.grab();
     overlay.show();
+    // Fill the screen with a single, safe geometry change. Drag-resize is
+    // disabled (it crashes CEF — see BrowserWindow creation), so maximise
+    // is how the window grows to fit the display. Desktop only: under
+    // gamescope the window is already born at the inner-X resolution and
+    // there's no WM to maximise against.
+    if (!gamescopeMode) {
+      try {
+        overlay.maximize();
+      } catch (e) {
+        console.warn("[overlay] maximize on open failed:", e);
+      }
+    }
     atoms.show().catch((e) => console.warn("[overlay] atoms.show:", e));
     state.isOpen = true;
     broadcastOverlayVisibility();
--- a/apps/loadout-overlay/src/bun/index.ts
+++ b/apps/loadout-overlay/src/bun/index.ts
@@ -252,16 +252,12 @@ const overlay = new BrowserWindow({
   title: "Loadout Overlay",
   frame: { x: 0, y: 0, width: OVERLAY_WIDTH, height: OVERLAY_HEIGHT },
   titleBarStyle: "default",
-  // Create the window NON-RESIZABLE. Live drag-resize crashes the overlay:
-  // dragging a corner (or dragging the window between monitors) streams a
-  // storm of geometry changes that reallocate CEF's window surface faster
-  // than it can keep up → SIGSEGV/SIGABRT with non-deterministic crash
-  // sites (heap corruption). A SINGLE atomic geometry change is safe — the
-  // WM "maximise" button never crashed in testing — so we disable user
-  // resize and instead maximise on open (see toggleOverlay). This overrides
-  // Electrobun's hardcoded `Resizable: true` default via the styleMask
-  // spread in BrowserWindow's constructor.
-  styleMask: { Resizable: false },
+  // The window is freely resizable. The drag-resize / drag-between-monitors
+  // crash was an Xlib thread-safety bug in Electrobun's native wrapper
+  // (OnPaint painted the OSR buffer on the CEF UI thread while
+  // process_x11_events drained events on the main thread, same Display, no
+  // XInitThreads) — fixed in our patched libNativeWrapper.so (electrobun
+  // #426). See apps/loadout-overlay/vendor/README.md.
   transparent: false,
   hidden: !DESKTOP_SMOKE_TEST,
   url: process.env.ELECTROBUN_DEV_URL ?? "views://overlay/index.html",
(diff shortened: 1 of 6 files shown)


srsholmes:
## Investigation log — overlay crash on window move/resize (desktop mode)

Context: while testing a fresh install on a multi-monitor desktop, the overlay (`loadout-overlay.service`) **crash-loops** when the window is interacted with. This PR started as "re-enable GL" but the GL change turned out **not** to be the fix — full findings below so we can resume.

### Symptom
- **Drag-resize from a corner** → crash (SIGSEGV/SIGABRT)
- **Drag the window between monitors** → crash
- **WM "maximise" button** (single atomic resize) → ✅ fine
- **Idle** → ✅ stable (survives many minutes untouched; only *interaction* triggers it)
- systemd `Restart=on-failure` then crash-loops it (~22 MB core each time)

### Crash signature
- `status=139` (SIGSEGV) in the main `bun` process; also batches of `SIGABRT` in `bun Helper` (CEF) subprocesses
- **Crash sites differ run-to-run, mix of SIGSEGV + SIGABRT → heap corruption**, not a clean single fault
- Crash surfaces in Bun worker threads, but the corruption source is the native CEF/GTK host layer
- CEF/Chromium **145.0.7632.68**, Electrobun **1.16.0**

### Hypotheses tested and RULED OUT
1. ❌ **`GDK_GL=disable` / software-render surface realloc** (the PR #113 hazard). Removed `GDK_GL=disable` to re-enable GTK GL (commit 1b9533a) → still crashes. `GDK_GL` only affects the GTK *host* window; CEF renders via its own path.
2. ❌ **Our libxcb FFI property watcher** (`x11.ts` `pollPropertyChanges`). Ran with `OVERLAY_FORCE_XPROP=1` (bypasses libxcb, uses xprop subprocess) → still crashes on move. Our xcb event parser is not the cause.
3. ❌ **CEF software rendering** (Electrobun's default `--disable-gpu` / SwiftShader). Set `disable-gpu:false` (verified real GPU: `--use-gl=disabled` dropped from the GPU process) → still crashes.
4. ❌ **device-scale-factor change crossing displays.** Set `--force-device-scale-factor=2` (verified applied) → still crashes.

### Partial mitigation (committed as WIP, commit 0239429 — not what we want long-term)
- **Non-resizable window + maximise-on-open** (`styleMask: { Resizable: false }`). **Stops the corner-resize crash**, but **not** the cross-monitor move crash, and it removes resizability (we want users to be able to drag/resize freely). Revert if pursuing free resize.

### Environment (very relevant)
Dual monitor, radically mismatched — moving between them forces CEF to re-create its render surfaces:

| Display | Resolution | ~DPI | Refresh | Notes |
|---|---|---|---|---|
| eDP-1 (handheld) | 2560×1440 | ~430 | 144 Hz | rotated (portrait-native) |
| DP-3 (external) | 3840×2160 | ~157 | 60 Hz | |

Huge DPI **+** refresh-rate **+** rotation deltas. Scale-factor pinning alone didn't fix it, so refresh/rotation/color-profile changes are likely also forcing surface re-creation.

### Leading theory
Heap corruption in **Electrobun's native CEF/GTK host (`libNativeWrapper.so`)** when CEF re-creates its rendering surface on a geometry/display-property change. A single atomic change (maximise) survives; a **resize storm** (corner drag) or a **display change** (crossing monitors) corrupts. Multi-monitor move crashes are a known Chromium/CEF class: chromium #422045260, hyprwm/Hyprland #11957/#11961, chromiumembedded/java-cef #477.

### Next steps (resume here)
- [ ] **Single-monitor isolation:** on the handheld panel alone, does *free resize* still crash? (Earlier: corner-resize crashed single-screen; maximise was fine.) Separates "resize storm" from "display change."
- [ ] Pin more display invariants together and retest the move: `force-device-scale-factor` + `--force-color-profile=srgb` + try matching refresh; also test moving between two monitors with *identical* settings.
- [ ] Dig into `libNativeWrapper.so`'s ConfigureNotify / display-change handler — we already inject a patched wrapper via `scripts/inject-patched-wrapper.sh` (CPU-spin fix), so the capability to patch it exists. Look for a surface-realloc race / use-after-free.
- [ ] File upstream with Electrobun (CEF 145, multi-DPI move repro). Consider debug-symbol build of the wrapper to get a real backtrace (current cores are stripped → `bun + 0x…` only).
- [ ] Ship-mitigation fallback if root cause is upstream: coalesce/debounce resize events, or constrain the window to one output.

### Machine state (local, NOT in repo)
- Live experiment flags reverted (`build.json` clean; `GDK_GL` removed from the installed unit via live edit).
- Overlay service **stopped** to avoid overnight core-dump spam — restart with `systemctl --user start loadout-overlay`.
- Coredumps preserved (`coredumpctl list`) for later symbolication.

srsholmes:
✅ **Validated and working on-device.** Resize from corners and cross-monitor dragging no longer crash with the patched wrapper deployed (real overlay, `NRestarts=0`). Root cause was the multithreaded-Xlib race in Electrobun's native wrapper (missing `XInitThreads()`) + the unthrottled ConfigureNotify storm — both fixed in the vendored `libNativeWrapper.so`.

Issue #140, opened 2026-06-21
Add timesplitters remake


srsholmes:
https://www.timesplittersrewind.com/

Add in timesplitters remake to recomp

PR #141, opened 2026-06-21
feat(recomp): add TimeSplitters Rewind to the catalog

srsholmes:
Issue #140. TimeSplitters Rewind is a fan-made Unreal Engine remake of
the trilogy — not strictly a recompilation, but it belongs in the same
catalog line as the recomps and runs the same way (Windows build via
Proton, registered as a non-Steam shortcut).

It ships only on ModDB/IndieDB, not GitHub, so the prebuilt download
path needed two small, general-purpose hooks on a catalog entry:

  - downloadHosts: extra hostnames the download's final (post-redirect)
    URL may resolve to, merged on top of the GitHub object-CDN defaults
    so downloadFile's host gate doesn't refuse the legitimate mirror.
  - downloadFilename: the local filename to save under, since an IndieDB
    mirror URL ends in an extension-less hash token and extractArchive
    dispatches the unpacker by file extension.

Both decisions are factored into pure, exported helpers
(resolveDownloadFilename / resolveDownloadHosts) with unit tests. The
entry pre-resolves the v0.3 IndieDB mirror via latestAssetUrl and pins
the DBolical hosts; repo is left empty so update checks skip it cleanly
(backend gates live lookups on a non-empty repo).

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_011T3LK8ovVPiGRprFjEruDa


srsholmes:
Recreated as #146 after a git history rewrite (MAC-address scrub) left this PR un-reopenable. Continue review there.

PR #154, opened 2026-06-23
docs(readme): format FAQ questions as headings

srsholmes:
Follow-up to #153. Makes each FAQ entry easier to scan:

- Each question is now an `###` heading (slightly larger than the surrounding body text) instead of bold inline text.
- A blank line now separates each question from its answer.

Pure formatting — no wording changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,19 +130,22 @@ entirely.
 
 ## FAQ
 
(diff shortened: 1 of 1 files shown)


srsholmes:
Pushed directly to main.

PR #190, opened 2026-07-01
feat(steamgriddb): art up owned-but-not-installed games

srsholmes:
The SGDB picker only listed installed games (appmanifest_*.acf) plus
non-Steam shortcuts, so games a user owns but hasn't downloaded never
appeared and couldn't be given custom artwork — even though the apply
path (SetCustomArtworkForApp + grid-folder write) works for any appId.

Add an opt-in owned-library source:

- @loadout/game-library: `scanLibrary` gains an `ownedApps` option;
  when passed, owned apps with no on-disk manifest are synthesized into
  the list (sizeOnDisk 0, same CDN + local art URLs as installed apps,
  inserted before the collection merge so favorites still get tagged).
  Art-URL building is extracted into a shared `buildSteamArtUrls` helper
  so installed and synthesized entries stay byte-identical.
- __core:game-library service: new `getFullLibrary()` reads the full
  owned library from Steam's in-memory appStore over CDP (getAllApps,
  same pattern protondb-badges uses) and merges via `ownedApps`. Falls
  back to installed-only with `ownedAvailable:false` when Steam is
  closed. Cached separately so the shared installed-only `getGames`
  every other plugin relies on is untouched.
- steamgriddb UI: consume `getFullLibrary`, badge not-installed tiles,
  and show a "start Steam" hint when the owned library is unavailable.

Tests cover the owned merge (synthesis, no-overwrite, collection tag,
malformed-entry guard, no-op regression), the service fallback/caching,
and the plugin badge + apply-with-source path.

Claude-Session: https://claude.ai/code/session_015XBH4Pu55D76rvfVZTzTQD

### DIFF
--- a/apps/loadout/src/loader/services/game-library.test.ts
+++ b/apps/loadout/src/loader/services/game-library.test.ts
@@ -13,9 +13,16 @@ import type { GameInfo, GameLibraryChangedEvent } from "@loadout/types";
 // caching and broadcast wiring lives in the loader.
 let mockGames: GameInfo[] = [];
 let scanCalls = 0;
+/** Records the options the service passed to the most recent
+ *  `scanLibrary` call — lets `getFullLibrary` tests assert the owned
+ *  apps are threaded through. */
+let lastScanOpts: { ownedApps?: Array<{ appId: string; name: string }> } | undefined;
 mock.module("@loadout/game-library", () => ({
-  scanLibrary: async () => {
+  scanLibrary: async (
+    opts?: { ownedApps?: Array<{ appId: string; name: string }> },
+  ) => {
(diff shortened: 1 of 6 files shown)


srsholmes:
Parking this: opening the plugin crashes Steam's webhelper via the SharedJSContext IPC collision described in #191. The goal (owned-but-not-installed games) stands — tracked in #191 for a collision-safe implementation (reuse the injector connection or read from a non-CDP source). Reverting to stable main meanwhile.

srsholmes:
closing. see [comment](https://github.com/srsholmes/loadout/pull/190#issuecomment-4856553101). Might come back later.

PR #215, opened 2026-07-17
feat(wifi): recover a crashed WiFi radio without a reboot

srsholmes:
## What & why

On the OneXPlayer Apex (Intel AX210, `iwlwifi`) the WiFi firmware crashed in the field — dmesg showed `HW problem - can not stop rx aggregation`, `failed to remove key ... (-5)` and an endless firmware reload loop, leaving NetworkManager's wifi device stuck `unavailable` with no way back but a reboot. The fix validated live on the device was a plain driver reload (`modprobe -r iwlmvm iwlwifi && modprobe iwlwifi`) — NetworkManager then reconnects the saved network on its own.

This adds that recovery to the **wifi** plugin:

- **"Recover WiFi radio" button** — reloads the driver in place, escalating to a PCI function reset and then a PCI remove+rescan when the reload isn't enough. Both escalation tiers only ever run when the radio is already dead, so the worst case equals the status quo (a reboot was needed anyway).
- **"Auto-recover radio" toggle** (off by default, persisted) — a 12s watchdog that fires only when the radio is `unavailable` *and* rfkill / the NM radio switch say it shouldn't be, so it never fights a deliberate off. Debounced (2 consecutive polls), cooled down (60s between attempts), and it suspends itself after 3 consecutive failed recoveries so it can't modprobe-crash-loop.

Gotchas from the live debugging, encoded and regression-tested:

- The interface can come back **renamed** after a reload (`wlan0` → `wlan1`) — the wifi device is re-detected from `nmcli` at every step, never hardcoded.
- `iwlmvm` must unload before `iwlwifi` (else `FATAL: Module iwlwifi is in use`); unmapped drivers get a generic `/proc/modules` "used by" holders fallback.
- The driver + PCI address are captured while the radio is healthy and persisted in plugin storage, so recovery still works after the interface has **vanished entirely**.

Post-review hardening (two multi-agent review rounds, findings in comments): the stored driver/PCI address are **re-validated on readback** before reaching `modprobe` argv or sysfs write paths; sysfs PCI writes are **timeout-bounded** so a kernel hang can't wedge the single-flight guard; `recover()` never rejects; manual precheck refusals don't poison the watchdog's suspension counter (watchdog-fired ones still count); the paused-banner only shows while auto-recover is enabled; and a pre-existing lost-update race on `wifi.json` (bare read+write vs the new serialized writers) was fixed.

New pure, dependency-injected orchestration in `plugins/wifi/lib/recovery.ts` (mirrors `powersave.ts`); backend wires it up with a single-flight guard so the button and watchdog share one run. Manifest gains the `modprobe` command permission. Plugin README + root gallery updated.

## How it was tested

- The underlying recovery recipe was validated live on the affected Apex (SteamOS, `iwlwifi`/AX210): module reload brought the crashed radio back without a reboot, including the `wlan0` → `wlan1` rename case; the shipped button was then exercised on-device (visible in the refreshed screenshot's `last recovery: ok` line).
- 94 wifi-plugin tests (69 recovery-lib/backend incl. the rename regression, holders retry, all three escalation tiers, watchdog debounce/cooldown/suspension, firing-tick path, nmcli-failure paths, storage readback validation; 10 UI specs incl. busy-label transition and error-notification path; plus the pre-existing powersave suite).
- Full repo suites: `bun run test:backend` (the one `sound-loader` failure is pre-existing on a clean tree — device-local EACCES on the Steam dir), `bun run test:ui` all green.
- Two rounds of multi-agent review (correctness / root-daemon safety / tests / maintainer docs); all verified findings fixed in follow-up commits.

## Checklist

- [x] `bun run typecheck` passes (0 errors)
- [x] `bun run lint` passes (0 errors)
- [x] `bun run test` passes
- [x] Added/updated tests for new backend/lib behaviour
- [x] Updated docs / `NOTICE` / `THIRD_PARTY_LICENSES.md` if this wraps or bundles third-party code


srsholmes:
## Multi-agent review summary

Three parallel reviewers (correctness, root-daemon safety, tests/conventions) audited this PR; fafa500 applies everything actionable.

**Real bugs found & fixed**
- **Lost-update race on `wifi.json`** — `setPowerSaveDisabled` still used a bare read+write while this PR added `mutatePluginStorage` writers to the same file; flipping the power-save toggle during the onLoad driver capture could erase `lastKnownDriver` from disk (breaking future vanished-interface recovery). Now serialized.
- **Watchdog suspension poisoning** — three instant precheck refusals (e.g. pressing the button while rfkill-blocked) counted as failed recoveries and suspended the watchdog before it ever attempted a reload. Precheck refusals no longer feed the failure counter.
- **Stale "Auto-recovery paused" banner** shown while auto-recover was toggled off.

**Hardening applied**
- `lastKnownDriver.driver` / `.pciAddress` are re-validated on storage readback before reaching `modprobe` argv or the sysfs `reset`/`remove` paths (the reviewer's severity call: not a new privilege boundary given the user already owns the plugin source dir, but a data file steering root writes is a defense-in-depth regression worth closing).
- Sysfs PCI writes are timeout-bounded so a kernel-level hang can't wedge the single-flight guard (and the watchdog) forever; `recover()` now genuinely never rejects.
- Watchdog tick re-entrancy guard.

**Verified clean by the reviewers** (traced, no defect): single-flight guard ordering, reducer debounce/cooldown arithmetic, interface-rename handling, nmcli terse parsing incl. escaped colons and "connected (externally)", holders-retry filtering, timer lifecycles, minimal permission delta (`modprobe` only), and repo-convention consistency across lib/backend/UI.

**Coverage added for the flagged gaps**: firing watchdog tick (source `watchdog`), tick-during-manual early return, precheck-vs-real failure counting, `refreshLastKnownDriver` skip-when-unchanged, nmcli failure paths, invalid stored driver/PCI address, never-rejects contract, and the failed-recovery error notification.

Suites after the fix pass: 68 wifi tests green (`test.ts` + `spec.tsx` runners), repo typecheck and lint clean.

srsholmes:
## Review round 2 — findings & fixes

Three fresh reviewers on the post-hardening state. Fixed in `939777e` + `cdf94d9` + `644ee27`:

**Real bugs**
- **Recovery re-enabled WiFi power saving** (HIGH) — the reloaded driver's netdev comes back at the kernel default (power-save ON); the wake listener only re-asserts on resume, and on iwd-backed SteamOS the runtime `iw` layer is the only live one. A successful recovery therefore handed the power-save-off user straight back the dropout bug this plugin exists to fix. `recoverRadio` now re-asserts `power_save off` when the toggle is enabled.
- **Watchdog fire→refuse loop** — my round-1 fix exempting precheck refusals from the failure counter was too broad: a *watchdog*-fired refusal (dead radio, unresolvable driver) escaped suspension and would retry every cooldown forever. Only manual refusals are exempt now.
- **Unkillable modprobe wedge** — `modprobe` in uninterruptible D-state on wedged firmware survives SIGKILL, so the exec timeout never fires; the ladder's modprobe calls now race an outer deadline and abandon the process (same pattern as the bounded sysfs writes).
- **False instruction in the precheck error** — "open this plugin once while WiFi works" never captured anything; `getStatus` now fires the quiet skip-if-unchanged capture, and the card shows the captured driver (or a hint when none is).
- **UI spinner stick** — an in-flight `getStatus` snapshot taken mid-recovery could resolve after the terminal refresh and pin `recovering: true` forever; refreshes now drop out-of-order responses.

**Verified clean, notably:** every plugin backend loads at daemon start, so the persisted auto-recover toggle works after reboot without opening the UI (no `loadOnStartup` needed); the sysfs-timeout race can't produce an unhandled rejection (checked empirically on Bun 1.3.14, plus the loader's log-only rejection handler); event plumbing shapes; no apex-plugin interference (disjoint subsystems); next-gen Intel `iwlmld` opmodes are covered by the generic holders fallback.

**Accepted as-is (with rationale):** a permanently-absent device (BIOS-disabled WiFi) replays up to 3 recovery ladders per boot before suspending — suspension is process-local by design, and persisting it risks never retrying after genuine hardware changes; tier-1 unload failure aborts rather than escalating (deliberate: don't PCI-reset a still-bound device); `onUnload` doesn't await an in-flight ladder (self-heals via the next boot's watchdog).

Docs round: plugin README gained a Radio recovery section, root gallery regenerated, PR body counts refreshed. Suites: 72 wifi tests green, repo typecheck + lint clean.

PR #216, opened 2026-07-18
feat(tdp-control): target-FPS auto-TDP per game

srsholmes:
Add a per-game "Target FPS" mode to the TDP plugin. When enabled for a
game, a closed-loop controller nudges wattage up/down to hold the chosen
frame rate, using MangoHud as the live FPS source. Games can still use a
fixed wattage (existing behaviour); each game picks its own mode.

- lib/fps-controller.ts: pure control math — proportional step with an
  asymmetric deadband, one-tick settle skip, reduce-side hysteresis, and
  saturation/floor detection. No I/O or timers, fully unit-tested.
- lib/fps-reader.ts: tails the newest MangoHud CSV, incremental byte-offset
  reads, header-driven fps column, rolling average, stale/rotation handling
  (follows the no-.exists() pseudo-file read rule).
- lib/mangohud.ts: per-app log dir + the three keyed launch-option tokens
  (MANGOHUD_CONFIG / MANGOHUD=1 / mangohud) injected and removed
  individually so nothing is orphaned on disable.
- lib/tdp-profiles.ts: additive mode/targetFps/min/maxWatts fields, v1->v2
  store migration, setTargetFpsProfile, and applyManagedTdp so controller
  writes share the existing 250ms debounced/serialised commit queue (the
  documented SMU-mailbox / APU-hang mitigation).
- backend.ts: 2s controller loop, AC-aware watt bounds, start/stop wired to
  game launch/exit + the per-game toggle, FPS-source grace period, and RPCs
  (setGameTargetFps / clearGameTargetFps / getControllerState /
  enable|disable|getMangoHudStatus). A manual setTdp pauses the controller.
- app.tsx: per-game Fixed | Target FPS selector, target stepper, live
  fps/watts/status readout, MangoHud toggle, and target-mode grid badges.

Note: SteamOS's native handheld TDP is a manual slider + framerate *cap*;
it has no closed-loop "raise power to hit a target" mode, so this is
additive. The controller deliberately doesn't fight the manual slider.

Claude-Session: https://claude.ai/code/session_01UCN3xL3tW5vRECAR8qTnrY

### DIFF
--- a/plugins/tdp-control/app.tsx
+++ b/plugins/tdp-control/app.tsx
@@ -76,7 +76,49 @@ interface SavedGameProfile {
   appId: number;
   gameName: string;
   tdpWatts: number;
+  mode?: "fixed" | "targetFps";
+  targetFps?: number;
+  minWatts?: number;
+  maxWatts?: number;
 }
+
+/** Live snapshot of the closed-loop FPS controller (getControllerState). */
+interface ControllerState {
+  running: boolean;
+  appId: number | null;
+  currentFps: number | null;
+  targetFps: number | null;
+  currentWatts: number | null;
+  minWatts: number | null;
+  maxWatts: number | null;
+  settled: boolean;
+  reason: string;
+  mangoHudActive: boolean;
+}
+
+interface MangoHudStatus {
+  installed: boolean;
+  configured: boolean;
+  logging: boolean;
+}
+
+const TARGET_FPS_MIN = 30;
+const TARGET_FPS_MAX = 120;
+const TARGET_FPS_STEP = 5;
+const DEFAULT_TARGET_FPS = 60;
+
+/** Human labels for controller reasons (badge text). */
+const CONTROLLER_REASON_LABELS: Record<string, string> = {
+  "warming-up": "Starting…",
+  settling: "Adjusting…",
+  holding: "Holding",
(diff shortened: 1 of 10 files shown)


srsholmes:
Closing in favour of the approach that landed in #218.

The core problem this PR was compensating for turned out to be CPU boost filling the TDP envelope (draw ≈ limit regardless of workload). With #218 merged, the boost policy is re-asserted with every TDP write, so the limit is now a genuine ceiling: the user can cap the framerate, set a generous max TDP (e.g. 45–50 W), and the hardware idles down to what the workload actually needs — verified on-device (~30 W draw under a 30 fps cap with a 43 W limit).

That passive behaviour covers the main use case of this closed-loop controller with zero moving parts, whereas this PR adds a MangoHud dependency, CSV tailing, a 2 s control loop, and a profile-store migration. The residual wins (race-to-idle efficiency at a tightened limit, transient spike clamping, cap-free VRR play) don't justify that complexity. Our direction is to let the OS/firmware do the heavy lifting rather than run our own control loop on top of it.

If a meaningful battery delta between "high ceiling + cap" and "manually tightened limit" shows up in practice, we can revisit — but as a simpler design that builds on #218 rather than this branch.

PR #224, opened 2026-07-23
feat(battery): charge limit + bypass charging, rename to "Battery"

### LINKED ISSUE #157: Add in charge limit
Add in charge limit for devices

### LINKED ISSUE #158: add in charging bypass for battery
add in charing by pass for battery

srsholmes:
Closes #157, closes #158.

Adds two battery-health controls to the plugin (renamed from **Battery Tracker** → **Battery**). Both use the kernel's generic `power_supply` sysfs interface and let the vendor driver do the EC work, so they run on **any device whose kernel exposes the attributes** rather than needing per-device code.

## What's new

- **Charge limit (#157)** — cap charging at a set percentage (50–95%) to reduce long-term battery wear. Writes `charge_control_end_threshold` on the *discovered* battery (`BAT0`/`BAT1`/`BATT`). Broadly supported and reliable.
- **Bypass charging (#158)** — run from the adapter without charging the pack (less heat, no charge cycles when docked). Writes `charge_behaviour` (`auto`/`inhibit-charge`/`inhibit-charge-awake`), falling back to the legacy OneXPlayer `charge_type` attribute (DMI-gated to `ONE-NETBOOK`).

## Behaviour

- **Capability-driven UI** — controls only appear on devices that expose the relevant attribute.
- **Bypass works where the kernel driver actually enforces it.** Some devices expose `charge_behaviour` but their firmware/driver silently ignore the inhibit (the write succeeds and reads back as set). To avoid a control that looks like it worked while charging continues, the plugin **verifies by outcome**: after enabling "always" bypass, if the battery is still charging a few seconds later, it shows a warning banner. This is device-agnostic (no per-model logic) and scoped to "always" only — "while awake" legitimately keeps charging until the device sleeps.
- **Persistence & reapply** — settings saved via `@loadout/plugin-storage`, rewritten at every service start (firmware forgets thresholds across reboots). Nothing is written on startup when the setting is "disabled", so another charging tool isn't overridden.
- **Optimistic UI** — after a successful write the UI reflects the value written rather than immediately re-reading (EC-backed attributes can lag a read-back and would otherwise snap the control back).
- Writes use the root-backend `Bun.write` + `tee` fallback pattern (added `tee` to plugin permissions).

## Cleanup

- Removes the dead, `BAT0`-hardcoded `setChargeLimit`/`getChargeLimit` RPCs from `tdp-control` so the **Battery** plugin is the single owner.
- Also fixes the shared `Select` dropdown (portal + UI-scale aware) so its menu can't be clipped by a card or render at the wrong scale — benefits every plugin that uses it.

## Device support

| Family | Charge limit | Bypass |
|---|---|---|
| ROG Ally / Ally X | ✅ | if the kernel driver enforces `charge_behaviour` |
| OneXPlayer / AOKZOE | ✅ | driver-dependent — some models/firmware expose it but don't enforce it (the warning banner covers this) |
| MSI Claw / AYANEO / GPD | if the kernel driver exposes the attrs | same |
| Steam Deck | ❌ (no kernel interface) | ❌ |

Note: charge limit is the broadly-reliable win. Bypass depends entirely on the device's kernel driver + firmware honouring the inhibit — on hardware that exposes but ignores it, the plugin now says so rather than failing silently.

## Tests

- Pure-helper tests, backend charge-control tests (detection incl. the force-discharge false-positive guard, restore-on-load incl. don't-clobber-when-disabled, all RPCs, DMI gate), and frontend coverage including the bypass effectiveness-check (warns on always+charging; does not warn on while-awake).
- `typecheck`, `lint`, `knip`, `check:specs` pass.

## Reviewer notes

- Display name is now "Battery"; the internal id `battery-tracker` is intentionally unchanged (RPC routes, storage file, overlay status bar).


srsholmes:
## Field note: bypass charging is a no-op on the OneXPlayer APEX (firmware/driver gap, not a plugin defect)

Documenting an investigation done on a **OneXPlayer APEX** (Strix Halo, `ONE-NETBOOK`/`ONEXPLAYER APEX`, **BIOS 0.14**, kernel 6.18, `oxpec` driver) so reviewers/future-me have the context behind the effectiveness-check and the honest device-support wording.

### TL;DR
The plugin writes `charge_behaviour` **correctly**, but this device's firmware/driver silently ignores the inhibit. **Charge limit works; bypass does not** — and no plugin change can fix that, because the gap is below us in the kernel driver + EC firmware. The new effectiveness-check exists precisely to surface this instead of shipping a silent dud.

### What we confirmed (read-only EC dumps via `ec_sys`)
`oxpec` uses EC reg `0xA3` = charge limit % and `0xA4` = charge inhibit (`auto=0x00`, `inhibit-charge-awake=0x01`, `inhibit-charge=0x03`). Diffing full 256-byte EC dumps across all three bypass modes:
- `0xA4` is the **only** mode-correlated byte, and `oxpec`'s writes **land correctly** (`00`/`01`/`03`).
- …yet the battery keeps `Charging` in every mode (capacity climbed 91→94% with `inhibit-charge` active).
- `0xA3` (charge limit) **is** honoured — set 80, it sticks and enforces.

So the APEX EC *accepts* the `0xA4` write but doesn't *act* on it. `charge_behaviour` being exposed ≠ enforced — and the probe can't tell the difference (the value reads back as set), which is why we verify by **outcome** instead.

### What we ruled out
- **Not a wrong value.** OneXConsole (the vendor Windows app) *can* bypass on this device, so the EC supports it. Its value for "always" looked like `0x0B`, but writing `0x0B` to `0xA4` while actively charging **also did nothing** — so it's not a simple value fix; the real path is likely a multi-register/native sequence.
- **Exact sequence not statically recoverable on Linux.** OneXConsole's privileged helper (`CompatLayerCT.exe`, .NET, WinRing0 EC access) is **code-protected** — every method body is a `throw` stub, restored at runtime by a native hook. Register *names* are readable (`EC_ADDR_BYPASS_POWER`, `EC_ADDR_POWER_SUPPLY_MODE`, `EC_ADDR_FORCE_CHARGE_MIN`) but the values/sequence are not.

### Likely real fixes (all outside this plugin)
- A **`oxpec` upstream change** for the APEX/`oxp_fly` entry, once the correct register/sequence is known.
- A **BIOS update** — BIOS 0.14 is very early, and OneXPlayer shipped BIOS updates that *added* this feature on the X1 Mini/Intel/F1. A future APEX BIOS could make `oxpec`'s existing standard path work with **zero plugin changes**.

### Why this doesn't block the PR
The plugin is generic and correct (standard `charge_behaviour`, capability-probed, no per-device code). Bypass works on any device whose kernel driver enforces the inhibit; the APEX is a confirmed non-enforcing case, and the effectiveness-check now tells the user that honestly rather than pretending it worked. Charge limit — the bigger battery-longevity win — works on the APEX today.

PR #225, opened 2026-07-23
feat(plugins): never load disabled plugins into the backend

srsholmes:
## Problem

Disabling a plugin was purely a **frontend visibility filter**. The backend loader (`plugin-manager.ts` → `loadPlugins`) bundled, dynamically imported, and ran `onLoad()` for **every** plugin directory on disk — as root — regardless of the UI toggle. So a "disabled" hardware plugin (`tdp-control`, `fan-control`, `rgb-control`, `input-plumber`) kept running and could conflict with tools like **Decky Loader** (e.g. two things driving ryzenadj / fan curves at once).

## Approach — split discovery from instantiation

A disabled plugin's code must never be imported or executed. That required an architectural shift:

- **`discoverPlugins()`** scans manifests only (never code). **`loadPlugin()`** bundles + imports + `onLoad`s a single plugin. **`loadPlugins()`** composes them, **skips the disabled set**, and returns a registry tagging each plugin `loaded | disabled | error`.
- Config flips to a **`disabledPlugins` deny-list** (was the frontend-only `enabledPlugins` allow-list), migrated once at startup (`resolveDisabledPlugins`). This also fixes a latent bug — the old allow-list silently hid any newly-installed plugin; the deny-list defaults new plugins to **enabled**.
- The backend reads the disabled set at startup and **enforces** it.

## Runtime behaviour is asymmetric (no safe unload exists)

- **Enable** → the loader loads the plugin **live, no restart**, and broadcasts `plugins-changed` so the UI refreshes.
- **Disable** → code already ran and can't be torn down in place. Restarting the backend alone would sever the overlay's WebSocket, so a new **`restartApp()`** RPC bounces `loadout.service` **and** the `loadout-overlay` unit together (reusing the updater's `systemd-run` helper). The UI **prompts** the user to restart — welcome-flow prompt + Settings banner — rather than doing it silently. The overlay closes and reopens; the game keeps running.

## Closing the side doors

- `App.tsx` resolves `activePlugin` from `visiblePlugins`, so `#/plugin/<disabled-id>` / an `OpenPlugin` shortcut can't render a disabled plugin.
- App-bundle/asset routes **404** for non-loaded plugins; RPC to a disabled plugin returns a distinct **"is disabled"** error.
- New **`/api/plugins?installed=1`** lists every installed plugin with `status`, so Settings/Welcome can still list and re-enable disabled ones.
- The Steam injector needed **no change** — `?all=1` already reads the loaded map, which no longer contains disabled plugins.

## Out of scope

True per-plugin runtime unload / process isolation (already tracked as Audit A-009). Restart-on-disable is the honest interim guarantee that *disabled ⇒ not in memory*.

## Verification

- `typecheck` ✅ · `lint` ✅ (0 errors) · `knip` dead-code ✅
- Backend suite passes except one **pre-existing flaky** sound-loader timing test (fails identically on clean `main`).
- UI suite: 370 pass.
- New tests: config migration + new-plugin-defaults-enabled (`user-config-plugins.test.ts`), discovery / skip-disabled (`plugin-discovery.test.ts`), disabled-vs-missing RPC error (`rpc-handler.test.ts`).

> ⚠️ The restart flow uses installed `systemctl` / `systemd-run` units, so it only fully exercises on a real device install (not `vite dev`).

https://claude.ai/code/session_0142SqYAbAEV48i7dFL7YDgS


srsholmes:
## Automated review pass (3 agents: backend, overlay/UX, adversarial-invariant)

**Verdict: the central invariant holds** — a disabled-at-startup plugin is never discovered-into-loaded, bundled, imported, `onLoad`-ed, RPC-reachable, injected into Steam CEF, broadcast to, or served over HTTP. The adversarial pass walked all 8 channels independently and found no high/medium holes. The migration (`enabledPlugins` allow-list → `disabledPlugins` deny-list) is exactly-once, correct set math, and free of the mutex re-entrancy deadlock trap (it calls the raw non-enqueuing helpers).

### Fixed in 27d4261
- **[Medium] keepAlive enforcement gap** — the keepAlive render layer filtered the full *loaded* list and keyed `active` off the raw route id, so a disabled keepAlive plugin (`browser`) could still render full-screen via an `OpenPlugin` shortcut / `#/plugin/<id>` deep link / last-tab restore. Now filtered through `isEnabled` (also unmounts its webview).
- **[Low] Welcome restart failure swallowed** — `restartApp()` errors (e.g. 409 update-in-progress) now toast instead of no-op'ing.
- **[Low] Welcome save failure ignored** — a failed `disabledPlugins` persist now toasts and skips the (pointless) restart prompt.
- **[Low] Perf** — `onUserConfigChanged` skips the registry enable-scan when the disabled set is unchanged, so it no longer runs on every unrelated config PATCH.

### Known limitations (acceptable-by-design, not fixed)
- **Live-enable doesn't re-inject Steam CEF targets** until the next Steam page reload — backend RPC + the overlay sidebar work immediately, but a plugin whose value is a CEF-injected Steam-menu entry won't appear there until reload.
- **A plugin that errored at load can't be re-enabled at runtime** — it needs an app restart to re-attempt the bundle (deliberate: retrying `Bun.build` on every config PATCH would be wasteful).
- **Pending-disable stays loaded until restart** — disabling an already-running plugin can't unload its code (no teardown path); the UI hides it immediately and prompts for the restart that actually clears it.

All checks green: typecheck, lint (0 errors), knip dead-code, backend loader tests, UI suite (370 pass). The one unrelated flaky sound-loader timing test fails identically on `main`.

srsholmes:
## Second review pass (3 agents: header-button, merge-integration, adversarial-UX)

Covered the new "Restart Loadout" header button, the first-round review fixes, and the `origin/main` (battery PR #224) merge.

**Merge integration: clean.** The enforcement files (`plugin-manager.ts`, `index.ts`, `routes/plugins.ts`, `rpc-handler.ts`, `user-config.ts`) came through the merge byte-for-byte; the `lastDisabledKey` early-out only gates the safe direction (runtime *enable*); the battery rename kept `plugin.id = "battery-tracker"` (all 25 plugins have dir==id); loader tests (160) + typecheck green.

**Header-button lifecycle: verified correct.** `pendingRestart` appears on disable-of-a-loaded-plugin and vanishes on revert (incl. partial reverts and startup-disabled enables); keepAlive fix unmounts a disabled keepAlive plugin; button can't get stuck disabled (self-heals on WS reconnect).

### Fixed in b74bea5
- **[HIGH] Restart raced the unflushed disable PATCH.** The Settings toggle persists fire-and-forget; the header button restarted immediately, so a lost race meant the backend re-read stale config, reloaded the just-disabled plugin, and `loadUserConfig` then overwrote the mirror — silently losing the disable with no affordance left. The button now `await`s `setConfigValueFlushed` (which also drains the queued toggle write via the FIFO config mutex) and aborts + toasts if it fails.
- **[MEDIUM] Rapid enable→disable double-tap could load a disabled plugin.** `onUserConfigChanged` captured a per-callback snapshot of the disabled set; the queued enable task now reads the *live* set, so a plugin re-disabled before the task runs is skipped (its `onLoad` never runs).
- **[low] WelcomeScreen RestartPrompt** gained a re-entrancy guard + "Restarting…" feedback (matches the header button).
- **[low] Welcome wizard** now uses the installed-plugins loading flag it actually renders from (no empty-list flash).

### Acceptable-by-design (unchanged)
- A disabled-pending-restart plugin stays fully loaded (RPC/broadcast/hardware) until restart — no teardown path exists; surfaced by the header button. The button is Settings-scoped per the feature request; leaving Settings hides the affordance while the plugin keeps running until restart. Flagged for product judgment; not changed.
- Errored plugins show as ON but need a restart to retry (runtime-enable only reloads `disabled`-status entries).

All checks green: typecheck, lint (0 errors), knip, loader tests (160), UI suite (378). Overlay redeployed to the test device with the flush fix; the backend MEDIUM fix needs a `bun run install-local` (sudo) to land on-device.

srsholmes:
## Third review pass (3 agents: footer/toast/hook, adversarial-UX, holistic merge-readiness)

**Verdict: GREEN — merge-ready.** No high/medium defects across all three. typecheck ✅, loader tests (160) ✅, UI suite (378) ✅, knip ✅.

Verified sound this round:
- **Invariant intact end-to-end** after all the churn — discovery/load split, `disabledIds` skip, migration, `isDisabled` RPC error, 404 for disabled bundles/assets, and disabled plugins never injected into Steam CEF (injector reads `?all=1` = loaded-only).
- **Footer restart + toast + `useRestartLoadout`** — no dead code from the removed header button, focus keys distinct, hook can't get stuck, footer state consistent across a real restart and partial reverts.
- **Flush-before-restart** genuinely correct (writes the full array + awaits; last-write-wins, no stale/unpersisted-disable restart).
- **Migration safe** for upgraders — faithful allow→deny conversion, no data loss.

### Fixed in 2af7518 (comments only, no behavior change)
- Two stale comments still said the restart button lives in the "Settings header" — it's the footer status bar now.
- The `index.ts` "live disabled set" comment overstated its guarantee: it's **best-effort**. A rapid enable→disable of a *startup-disabled* plugin can still load it once (the enable microtask reaches `loadPlugin` before the disable PATCH is persisted), landing it in the normal disclosed pending-disable state (loaded + "Restart required", cleared on restart). A hard guarantee needs a plugin unload path (Audit A-009). Comment now says so.
- The flush comment's "FIFO mutex" rationale was imprecise; corrected to full-array-write + await.

### Noted, not changed
- The disable toast shows a green success checkmark because `notify` has no warning/info kind (only success/error/loading). Semantically off but harmless; expanding the shared toast system wasn't warranted for a nit.
- Optional follow-up test coverage: the runtime-enable path and the disabled-route 404s.
- Release note for upgraders: a plugin newly bundled in this update that wasn't in an existing user's old `enabledPlugins` migrates to *disabled* (migration is faithful to the old allow-list); new plugins default-enabled only after migration.

All feature work deployed and verified live on-device across rounds; the backend `index.ts` change lands on-device via `bun run install-local`.

PR #228, opened 2026-07-27
fix(overlay): capture the controller on Steam Deck (hidraw nav + per-host input strategy)

srsholmes:
## Problem

On Steam Deck the overlay never captured the controller:

- Steam reads the built-in controller (and external pads) via **hidraw**, which `EVIOCGRAB` cannot intercept — Steam BPM / Steam-Input-fed games kept responding behind the open overlay.
- In BPM home there isn't even an evdev nav source: the virtual X360 pad (`28de:11ff`) only exists while a game runs, so the overlay opened with **zero nav** (`input intercept ready — 0 controller(s)`).
- The strategy was inferred from IP absence (`readVirtualPadsForNav = !ipHandle.available`) rather than platform detection.

## Fix

New pure `decideInputStrategy()` (`apps/loadout-overlay/src/bun/lib/input-strategy.ts`) picks per host:

| Host | Nav source | Capture | Freeze |
|---|---|---|---|
| IP-managed (OXP APEX…) | IP DBus stream | InterceptMode + grabs | ON (**unchanged**) |
| **Deck, watcher OK** | **our own hidraw read** | built-in nodes + virtual pad grab-only, Steam SIGSTOP in game mode | **ON** |
| Deck, watcher failed / `DECK_OVERLAY_DECK_CAPTURE=0` | virtual pad (legacy) | evdev grabs only | OFF (exact legacy) |
| other hosts | evdev read+grab | evdev grabs | OFF |

- `packages/deck-hid`: `decodeNavState()` — full 64-byte report decode (buttons, d-pad, sticks), offsets cross-validated against SDL `SteamDeckStatePacket_t` / InputPlumber `hid_report.rs` and the repo's Jupiter-verified `DECK_BUTTONS`.
- `deck-hidraw-watcher`: wake path untouched; new nav mode (`setNavActive`) emits NavController-shaped events with a silent first-frame baseline (held wake button can't fire into a fresh overlay).
- `input-intercept`: `injectNavEvents()` external-source entry with grab-generation isolation + 25 ms repeat pump; `grabDeckBuiltinNodes` makes every `28de:1205/1206` node grab-only (incl. mouse-only lizard nodes — **desktop-mode capture**, since desktop consumers read evdev; also prevents raw-gamepad-node double-fire when Steam isn't running).
- Steam freeze on Deck is gated on the hidraw watcher having actually opened — a failed watcher falls back wholesale to today's behavior.
- Steam/QAM are blocked as wake bindings (`d07a1af`): they're reserved system buttons that Steam sees live via hidraw before any freeze. The picker no longer offers them, `setWakeButton` rejects them, and a binding persisted before the block is treated as unbound. **Deck-only** — the IP capability picker is untouched.
- `isSteamDeck()` centralized in `@loadout/devices` (was private to the input-plumber plugin).

**Works in Steam menus (no game running), in-game, and desktop mode; external wireless pads keep the evdev grab+read path and are silenced behind the overlay by the freeze.** InputPlumber-managed hosts are behaviorally untouched.

Kill switches: `DECK_OVERLAY_DECK_CAPTURE=0` (whole Deck strategy), `DECK_OVERLAY_SUSPEND_STEAM=1/0` (freeze override, pre-existing).

## Hardening (`27fdbbb`, after the first review round)

1. **Resume-burst:** Steam's hidraw fd keeps the *first* ~64 reports after the SIGSTOP, so SIGCONT replayed the wake-release edge into the game. On a frozen-Deck close the evdev release is deferred until SIGCONT + 500 ms; reopen inside the window cancels + flushes. Steam-BPM-side replay is unblockable — paddle wake bindings (R4/L4/…) avoid it.
2. **Watcher death mid-session** left freeze+grab armed with zero nav. `isAlive()`/`onDeath` now force-close and exit so systemd restarts and startup re-derives the strategy.
3. **Guide+B wake on Deck** — combo wake detection keeps running on grab-only controller nodes under the Deck strategy only (IP hosts unchanged; a mirror would double-toggle).
4. **Right-stick jitter** — 0.10 dead-band before quantization (evdev gets this free from kernel fuzz filtering; raw hidraw must do its own).

## Verification

**Counts (measured, not estimated):** 8 commits. Across the 8 touched test files: **163 tests pass, 112 of which exist on `main` → 51 net-new.** (An earlier revision of this description said "66 new tests" — that was wrong.) `bun run typecheck` clean; `eslint` clean apart from 3 warnings in test files.

**On-device, OXP APEX / SteamOS (non-Deck, InputPlumber host) — built and installed from this branch:**

```
[overlay] ip intercept ready — 1 composite device(s)
[overlay] ip intercept ACTIVE — using InterceptMode + DBus nav
[overlay] input strategy: deckNav=OFF steam-freeze=ON readVirtualPads=false
          (isDeck=false, ipManaged=true, hidraw=false, env: suspend=auto, deckCapture=auto)
[overlay] input intercept ready — 0 controller(s), 4 keyboard(s), 4 qam device(s) (readVirtualPadsForNav=false)
```

Byte-identical device counts and freeze decision vs the pre-PR build on the same hardware. No `[deck-hidraw]` lines, no `deck-builtin(grab-only)` devices. Also verified live on that host: `findDeckHidrawPath()` → `null` in 2.7 ms, `isSteamDeck()` → `false`, and `decideInputStrategy` reduces to main's exact values both with IP up and with IP down.

**Not verified:** everything Deck-specific. The hidraw nav decode, the lizard-node grabs, the Steam freeze on Deck, and the resume-burst deferred release all need real Jupiter/Galileo hardware. That is where the risk in this PR lives.

## Known accepted (documented, not fixed)

- Overlay opened before startup discovery completes is touch-only until close/reopen (pre-existing on `main`; window slightly wider).
- A wake press inside the 600 ms toggle debounce can leak one nav edge.
- Steam's live reaction to the *opening* wake press for non-paddle bindings.
- The grab/release loops assume every `28de:11ff` node has controller capabilities. True in the field — verified on the APEX, which exposes exactly one `11ff` node (`Microsoft X-Box 360 pad 0`, controller caps). A hypothetical `11ff` keyboard-only node would now be grabbed.

### DIFF
--- a/apps/loadout-overlay/package.json
+++ b/apps/loadout-overlay/package.json
@@ -14,6 +14,7 @@
   "dependencies": {
     "@noriginmedia/norigin-spatial-navigation": "^3.0.0",
     "@loadout/deck-hid": "workspace:*",
+    "@loadout/devices": "workspace:*",
     "@loadout/exec": "workspace:*",
     "@loadout/plugin-storage": "workspace:*",
     "@loadout/types": "workspace:*",
--- a/apps/loadout-overlay/src/bun/index.ts
+++ b/apps/loadout-overlay/src/bun/index.ts
@@ -55,6 +55,7 @@ import { routeWake, isStartupWakePhantom } from "./lib/wake-routing";
 import { loadPersistedShortcuts } from "./lib/persisted-shortcuts";
 import { decideInputStrategy } from "./lib/input-strategy";
 import { isSteamDeck } from "@loadout/devices";
+import { isBindableDeckButton } from "@loadout/deck-hid";
 
 // ---- State ------------------------------------------------------------------
 // PendingFlags + the flag-lifecycle helpers live in lib/overlay-state.ts so
@@ -879,8 +880,14 @@ async function readDeckWakeBinding(): Promise<string | null> {
   const raw = s.wake?.selectedRaw ?? null;
   // Raw is either a synthetic "deck:<Button>" string we wrote here, or an
   // InputPlumber capability string written on a non-Deck host. Only the
-  // deck:* form is meaningful for the watcher.
-  if (raw && raw.startsWith("deck:")) return raw.slice(5);
+  // deck:* form is meaningful for the watcher — and only bindable buttons:
+  // Steam/Qam are reserved system buttons (the picker no longer offers
+  // them, but a binding persisted before they were blocked must not arm
+  // the watcher either).
+  if (raw && raw.startsWith("deck:")) {
+    const name = raw.slice(5);
+    return isBindableDeckButton(name) ? name : null;
+  }
   return null;
 }
 
--- a/apps/loadout-overlay/src/bun/index.ts
+++ b/apps/loadout-overlay/src/bun/index.ts
@@ -387,6 +387,17 @@ let deckNavActive = false;
 /** Controller id prefix for hidraw-injected nav events (see
  *  externalNavSources in input-intercept). */
 const DECK_NAV_SOURCE = "deck-hidraw";
+// Deck resume-burst mitigation state (see the close path in toggleOverlay):
+// whether THIS open actually SIGSTOPped Steam, and the deferred evdev
+// release that stays grabbed across SIGCONT + drain so the game can't see
+// Steam's replayed hidraw backlog.
+let steamFrozenThisOpen = false;
+let pendingReleaseTimer: ReturnType<typeof setTimeout> | null = null;
+/** How long after SIGCONT the evdev grabs are held so Steam's buffered
+ *  hidraw backlog (≤64 reports) drains into a grabbed virtual pad instead
+ *  of the game. Chosen > TOGGLE_DEBOUNCE_MS - 250 so a reopen can overlap
+ *  it — the open path cancels + flushes the deferred release. */
+const DECK_RESUME_DRAIN_MS = 500;
 
 // Startup safety net for the frozen-Steam risk: a previous overlay instance
 // that crashed or was SIGKILLed *while open* could have left Steam SIGSTOPped
@@ -458,11 +469,18 @@ function forceCloseOverlay(reason: string): void {
     clearTimeout(pendingResumeTimer.current);
     pendingResumeTimer.current = null;
   }
+  if (pendingReleaseTimer !== null) {
+    clearTimeout(pendingReleaseTimer);
+    pendingReleaseTimer = null;
+  }
   if (steamPid.current === null) steamPid.current = findSteamPid();
   if (steamPid.current !== null) resumeSteam(steamPid.current);
+  // Emergency path: release immediately (no resume-burst drain — a wedged
+  // renderer is worse than a replayed edge) and thaw.
   intercept.current?.release();
   ipIntercept.current?.release();
   deckHidraw.current?.setNavActive(false);
+  steamFrozenThisOpen = false;
   atoms.hide().catch((e) => console.warn("[freeze-watchdog] atoms.hide:", e));
   try {
     overlay.minimize();
@@ -508,7 +526,20 @@ function toggleOverlay(source: string) {
     stopFreezeWatchdog();
     overlay.minimize();
     atoms.hide().catch((e) => console.warn("[overlay] atoms.hide:", e));
-    intercept.current?.release();
+    // Deck strategy resume-burst mitigation: while Steam was SIGSTOPped its
+    // own hidraw fd kept buffering (the kernel ring keeps the FIRST ~64
+    // reports after the freeze — which contain the wake button's release
+    // edge and any early nav input — and drops the rest). On SIGCONT Steam
+    // drains that backlog and forwards it to the virtual pad; if our evdev
+    // grabs were already gone, the game would replay those stale edges on
+    // every overlay close. So on a frozen-Deck close the evdev release is
+    // DEFERRED until after SIGCONT + a drain window, keeping the virtual
+    // pad (and built-in nodes) grabbed while the burst flushes. Steam's own
+    // BPM reaction to the buffered wake release can't be blocked this way —
+    // paddle wake bindings (R4/L4/…) avoid it entirely since Steam ignores
+    // unmapped paddles. Non-Deck hosts keep the immediate release.
+    const deferRelease = deckNavActive && steamFrozenThisOpen;
+    if (!deferRelease) intercept.current?.release();
     ipIntercept.current?.release();
     // Stop decoding Deck hidraw nav. The wake-bit path stays live so the
     // wake button can re-open; the diff baseline resets on next open.
@@ -529,8 +560,19 @@ function toggleOverlay(source: string) {
       pendingResumeTimer.current = setTimeout(() => {
         pendingResumeTimer.current = null;
         resumeSteam(pid);
+        if (deferRelease) {
+          if (pendingReleaseTimer !== null) clearTimeout(pendingReleaseTimer);
+          pendingReleaseTimer = setTimeout(() => {
+            pendingReleaseTimer = null;
+            intercept.current?.release();
+          }, DECK_RESUME_DRAIN_MS);
+        }
       }, 250);
+    } else if (deferRelease) {
+      // No Steam pid found — nothing to drain; release now rather than never.
+      intercept.current?.release();
     }
+    steamFrozenThisOpen = false;
     state.isOpen = false;
     broadcastOverlayVisibility();
     trace(`[toggle] ${source} → MINIMIZE`);
@@ -544,11 +586,26 @@ function toggleOverlay(source: string) {
     // so freezing it just wedges Steam — the bug this gate fixes. The evdev
     // grab below still runs in both modes, so the controller overlay keeps
     // working on the desktop without touching Steam.
+    // A rapid close→open can land while the close path's deferred SIGCONT /
+    // deferred release timers are still pending. Cancel them (this open owns
+    // the freeze again), and if the evdev release was still deferred (grabs
+    // held from last session), flush it synchronously so grab() below runs
+    // a fresh generation with clean NavController/combo state.
+    if (pendingResumeTimer.current !== null) {
+      clearTimeout(pendingResumeTimer.current);
+      pendingResumeTimer.current = null;
+    }
+    if (pendingReleaseTimer !== null) {
+      clearTimeout(pendingReleaseTimer);
+      pendingReleaseTimer = null;
+      intercept.current?.release();
+    }
     if (suspendSteamEnabled && isGameModeActive()) {
       if (steamPid.current === null) steamPid.current = findSteamPid();
       if (steamPid.current !== null) {
         suspendSteam(steamPid.current);
         startFreezeWatchdog();
+        steamFrozenThisOpen = true;
       }
     } else if (suspendSteamEnabled) {
       trace("[toggle] desktop mode (no gamescope) — skipping Steam freeze");
@@ -707,6 +764,28 @@ void (async () => {
       // nav mode can't be active before the first grab() anyway.
       onNavEvents: (events) =>
         intercept.current?.injectNavEvents(DECK_NAV_SOURCE, events),
+      // Mid-session stream death (driver rebind, controller re-enumeration
+      // around suspend/resume): under the Deck strategy the watcher is the
+      // ONLY nav source AND the wake button, while the baked-in strategy
+      // would still freeze Steam + grab everything on the next open — the
+      // exact stranded-with-zero-nav state the startup gate exists to
+      // prevent. The strategy can't be re-derived live (intercept options
+      // are fixed at startInputIntercept), so exit and let systemd restart
+      // us: startup re-probes and picks the right strategy (watcher back →
+      // full Deck strategy; still dead → legacy fallback). The startup
+      // unconditional SIGCONT covers a death while frozen.
+      onDeath: () => {
+        console.error(
+          "[overlay] Deck hidraw watcher died mid-session — " +
+            (deckNavActive
+              ? "restarting overlay to re-derive the input strategy"
+              : "Deck wake button lost until next overlay restart"),
+        );
+        if (deckNavActive) {
+          if (state.isOpen) forceCloseOverlay("deck hidraw watcher died");
+          process.exit(1);
+        }
+      },
     });
     if (handle) {
       deckHidraw.current = handle;
@@ -722,7 +801,9 @@ void (async () => {
   const strategy = decideInputStrategy({
     isDeck,
     ipAvailable: !!ipHandle?.available,
-    deckHidrawOpened: deckHidraw.current !== null,
+    // isAlive guards the (tiny) window where the stream died between open
+    // and this decision — a dead watcher must not enable the freeze.
+    deckHidrawOpened: deckHidraw.current?.isAlive() ?? false,
     suspendEnv: SUSPEND_STEAM_ENV,
     deckCaptureEnv: DECK_CAPTURE_ENV,
   });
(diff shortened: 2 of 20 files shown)


srsholmes:
## Multi-agent review results (3 independent passes)

**Lens 1 — IP-host regressions:** verdict *preserved*. `decideInputStrategy` reduces exactly to the old formulas whenever `isDeck=false` or `ipAvailable=true` (all env-override combinations checked); intercept loop changes are bit-identical with the new flags off (the only theoretical divergence would need a 28de:11ff node without controller caps, which doesn't exist — now documented); DMI refactor logic-identical, no import cycles. Only deviation: boot-time ordering (deck watcher now arms after IP discovery; also fixes a latent race where the watcher could fire before wake-routing was seeded).

**Lens 2 — Deck-path correctness** and **Lens 3 — tests/sweep** found four real issues, all fixed in 27fdbbb:

1. **Resume-burst (blocker):** Steam's hidraw fd keeps the *first* ~64 reports after the SIGSTOP (kernel drops new reports when the ring fills), so SIGCONT replayed the wake-release edge — into the game, since our grabs were already gone. Fix: on a frozen-Deck close, the evdev release is deferred until SIGCONT + 500 ms so the burst drains into a still-grabbed virtual pad; reopen inside the window cancels + flushes. Steam-BPM-side replay is unblockable — paddle wake bindings (R4/L4/…) avoid it, noted in code.
2. **Watcher death mid-session** left freeze+grab armed with zero nav. Fix: `isAlive()`/`onDeath` on the watcher; under the Deck strategy a dead stream force-closes and exits so systemd restarts and startup re-derives the strategy.
3. **Guide+B wake regression on Deck** (grab-only virtual pad skipped wake detection). Fix: combo wake detection keeps running on grab-only controller nodes under the Deck strategy only — IP hosts unchanged (mirror would double-toggle).
4. **Right-stick jitter:** rest drift straddling a quantization boundary streamed axis RPCs at report rate. Fix: 0.10 dead-band before quantization (evdev gets this from kernel fuzz filtering; raw hidraw must do its own).

Also fixed: grab/release log labels + counts, stale `isSteamVirtual` doc comment; new tests for dead-band boundary alternation, `onDeath`/`isAlive`, and a production-shaped wake-button-is-a-nav-button sequence.

Known accepted (documented, not fixed): overlay open before startup discovery completes is touch-only until close/reopen (pre-existing on main, window slightly wider); wake press inside the 600 ms toggle debounce can leak one nav edge; Steam's live reaction to the *opening* wake press for non-paddle bindings; `bun.lock` absorbed stale 0.6.0→0.7.1 drift from main.

https://claude.ai/code/session_01QBqPgKKhLTmtLgCeYNYBix

srsholmes:
## Testing guide for non-Deck / InputPlumber handhelds (written for the agent verifying on the OXP APEX)

This PR adds a Steam-Deck-only input strategy. **On any non-Deck host — and on any host where InputPlumber has CompositeDevices — the intended behavior is BIT-IDENTICAL to `main` (v0.7.1).** Your job on the APEX is to confirm nothing regressed. This comment gives you the full context: what changed, why the APEX should be unaffected, exactly what to look for, and how to diagnose/mitigate if something is off.

### 1. What this PR does (background)

On Steam Deck, Steam Input reads the built-in controller via **hidraw**, which `EVIOCGRAB` cannot intercept, and in BPM home there is no evdev nav source at all (the virtual X360 pad `28de:11ff` only exists while a game runs). The fix: on the Deck the overlay decodes nav from its own hidraw read (`deck-hidraw-watcher` nav mode), makes all Deck built-in evdev nodes (`28de:1205/1206`) + the virtual pad grab-only, and SIGSTOPs Steam while the overlay is open in game mode. Steam/QAM buttons are blocked as wake bindings (reserved system buttons). None of this should engage on the APEX.

### 2. The gate chain that keeps the APEX on the legacy path

`decideInputStrategy()` in `apps/loadout-overlay/src/bun/lib/input-strategy.ts` enables the Deck strategy only when ALL of:
1. `isDeck` — DMI `product_name` contains Jupiter/Galileo, or `sys_vendor` contains Valve. APEX reports `ONE-NETBOOK` / `ONEXPLAYER APEX ...` → **false**.
2. `!ipAvailable` — the APEX runs InputPlumber with ≥1 CompositeDevice → `ipAvailable=true` → Deck strategy **disabled regardless of anything else**.
3. `deckHidrawOpened` — requires a hidraw node with USB VID 0x28de PID 0x1205/0x1206 interface 2. Does not exist on APEX → `startDeckHidrawWatcher` returns null → **false**.

With gate 2 false, the function reduces exactly to main's logic: `readVirtualPadsForNav = !ipAvailable` (= false on APEX), `suspendSteamEnabled = env override else ipAvailable` (= true on APEX). `DECK_OVERLAY_SUSPEND_STEAM=1/0` semantics unchanged. The new intercept options (`grabDeckBuiltinNodes`, `externalNavSources`) are only set under the Deck strategy — off on APEX, making `input-intercept.ts` device selection and grab/release loops identical to main.

### 3. Expected journal lines on the APEX (`journalctl --user -u loadout-overlay`)

After a restart you MUST see:
```
[overlay] ip intercept ready — N composite device(s)        # N ≥ 1
[overlay] ip intercept ACTIVE — using InterceptMode + DBus nav
[overlay] input strategy: deckNav=OFF steam-freeze=ON readVirtualPads=false (isDeck=false, ipManaged=true, hidraw=false, env: suspend=auto, deckCapture=auto)
[overlay] input intercept ready — X controller(s), Y keyboard(s), Z qam device(s) (readVirtualPadsForNav=false)
```
You must NOT see:
- any `[deck-hidraw]` line (`findDeckHidrawPath` returns null before opening anything)
- any device opened as `deck-builtin(grab-only)`
- `deckNav=ON` or `isDeck=true`
- on close: `released N device(s)` should count the same devices main counted (the log word changed from `controller(s)` to `device(s)` — cosmetic only)

The one intentional boot-time difference vs main: the deck-hidraw probe + strategy decision now run between IP discovery and `startInputIntercept`, so the evdev intercept arms a few ms later at boot. Steady-state behavior after startup is unchanged.

### 4. Functional checklist (all should behave exactly as v0.7.1)

1. **Wake button** (IP profile → KeyF16): opens the overlay; pressing again closes it. Boot-persistence: reboot, wake button still works (reloadPersistedProfile path untouched — only `isSteamDeck()` moved to `@loadout/devices`, logic identical, re-exported from `wake-trigger.ts` so `backend.ts` imports are unchanged).
2. **Nav**: D-pad/left stick navigates the overlay via IP DBus (`gdbus monitor` stream), A/B/X/Y act, right stick scrolls.
3. **Capture**: while the overlay is open, the game and Steam BPM behind it receive NOTHING (InterceptMode=3 + Steam SIGSTOP + evdev grabs). On close, input returns to the game immediately; the known bounded resume-burst (≤64 stale reports into BPM) is pre-existing, not from this PR. **The deferred-release change (resume-burst mitigation) is Deck-gated** (`deckNavActive && steamFrozenThisOpen`) — on APEX the evdev release stays immediate on close, same as main.
4. **Wake-button picker** (input-plumber plugin UI): still lists the IP capability buttons from `CompositeDevice.Capabilities`. The Steam/QAM reserved-button block ONLY filters the Deck's synthetic button list (`wake-trigger-deck.ts`, reached only when `isSteamDeck()` is true) — the IP picker path (`wake-trigger.ts` / `profile.ts` `buttonOptions`) is untouched. Press-to-capture, Restart button (+#147 reset-failed behavior), and Disable Controller Input plugin: all unchanged.
5. **Hotplug**: pair a BT pad mid-session → it's opened/grabbed as before; its Steam virtual mirror stays grab-only.
6. **Crash recovery**: `systemctl --user kill -s SIGKILL loadout-overlay` while open → on restart Steam is SIGCONT'd (startup unconditional thaw, unchanged).

### 5. If something IS broken on the APEX

- First read the `input strategy:` log line — it prints every input to the decision. If it shows anything other than `deckNav=OFF ... isDeck=false, ipManaged=true, hidraw=false`, capture the line and also `cat /sys/class/dmi/id/product_name /sys/class/dmi/id/sys_vendor` and `busctl tree --list org.shadowblip.InputPlumber | grep CompositeDevice`.
- `isDeck=true` on an APEX would mean a DMI false-positive (Valve in sys_vendor — should be impossible); even then `hidraw=false` keeps the strategy off. `ipManaged=false` would mean IP discovery failed (check `systemctl status inputplumber`) — that's an environment issue, not this PR, but it changes behavior to the legacy evdev path (virtual pads read for nav, no freeze), same as it would have on main.
- Kill switches (set in the overlay service environment): `DECK_OVERLAY_DECK_CAPTURE=0` disables the whole Deck strategy (a no-op on a correctly-detected APEX, but a blanket guard); `DECK_OVERLAY_SUSPEND_STEAM=1/0` force the freeze on/off exactly as on main.
- The only shared-code invariant worth knowing: the grab/release loops changed from `isController` to `isController || grabOnly`. On an IP host this diverges from main ONLY if a `28de:11ff` evdev node exists WITHOUT controller capabilities (never observed; 11ff is always the virtual X360 pad with full controller caps). If you see an unexpected grab in the journal, check `grep -A5 11ff /proc/bus/input/devices`.

### 6. Files changed that execute on the APEX (all verified behavior-preserving)

- `apps/loadout-overlay/src/bun/index.ts` — startup reordering + strategy call (reduces to old values), close-path deferred release (Deck-gated)
- `apps/loadout-overlay/src/bun/native/input-intercept.ts` — new options defaulting off; loop-condition change discussed above; grab-only wake-combo exception gated on `grabDeckBuiltinNodes` (off)
- `apps/loadout-overlay/src/bun/native/devices.ts` — new `isDeckBuiltin` field (computed, unused when the option is off)
- `plugins/input-plumber/lib/wake-trigger.ts` — `isSteamDeck()` now imported from `@loadout/devices` (logic identical: Jupiter/Galileo substring, Valve fallback, empty-on-unreadable)
- `packages/devices/src/dmi.ts`, `packages/deck-hid/*` — pure additions
- NOT executed on APEX: `deck-hidraw-watcher.ts` nav mode (watcher never starts), `wake-trigger-deck.ts` (isSteamDeck false), `input-strategy.ts` Deck rows

https://claude.ai/code/session_01QBqPgKKhLTmtLgCeYNYBix

srsholmes:
## Second review round — 2 independent passes, 4 fixes applied (`f29087d`)

Lenses chosen to cover what the first round didn't: the hidraw decode (untestable on the APEX this was developed on) and the freeze/grab-release state machine.

### 1. The report filter tested a byte that cannot discriminate — Medium, real

Every call site gated on `report[0] !== REPORT_ID_INPUT`. **Byte 0 is the protocol version and is `0x01` on every Valve in-report** — the type discriminator is byte 2 (`ucType`). I verified this against `drivers/hid/hid-steam.c` directly rather than taking the reviewer's word:

```c
if (size != 64 || data[0] != 1 || data[1] != 0) return 0;
switch (data[2]) { case ID_CONTROLLER_DECK_STATE: /* = 9 */ ... }
```

and the driver's own layout comment: `0-1 | always 0x01, 0x00`, `2 | type of message`.

So the filter was a constant. Any other 64-byte in-report — Steam Input drives feature reports on this endpoint concurrently — was decoded as button + stick state. On a real Deck: phantom presses or a stick snap into a freshly-opened overlay, or a spurious wake toggle while closed.

The header comment named this exact hazard ("the hidraw stream interleaves a few report types; only 0x01 carries button + axis state") while filtering on the byte that can't detect it. New `isDeckStateReport()` mirrors the kernel gate; all four sites use it.

**Pre-existing on `main`** for the wake path — but this PR widened the blast radius from one wake bit to full nav decode including all four axes, so it belongs here.

### 2. `onDeath` exited before `atoms.hide()` could run — High

`forceCloseOverlay` schedules an un-awaited `atoms.hide()`; `process.exit(1)` on the next line killed the process before its first `await` resumed. Steam was always safe (SIGCONT and the evdev release are synchronous), but the atom writes outlive our window: ROOT `STEAM_TOUCH_CLICK_MODE` stays at `TOUCH_MODE_OVERLAY` — gamescope routing touch to a window that no longer exists — plus Steam's BPM `STEAM_OVERLAY` / `STEAM_INPUT_FOCUS` left force-zeroed.

`shutdown()` in `lifecycle.ts` already awaits a *retried* `hide()` with the comment "Failure to clear the atom strands the user (no input until reboot)". This was the one exit path skipping it. `forceCloseOverlay` now returns the hide promise; `onDeath` waits on it, bounded by a 1 s timeout in case xprop wedges.

### 3. Fast reopen could skip the deferred-release flush — Medium

The open path only flushed when `pendingReleaseTimer` was non-null — but that handle is armed *inside* the resume callback, so a reopen within the 250 ms pre-SIGCONT window cancelled the resume while it was still `null`. The flush was skipped, then `doGrab()` early-returned on `intercepting`, reusing the previous generation's NavController/combo state → stuck-held direction or dead nav for that session. Release is now unconditional (`doRelease()` no-ops otherwise).

Needs an event-loop stall to reach today, since `TOGGLE_DEBOUNCE_MS`(600) > 250 — but that guard was implicit. The existing comment justifies `DECK_RESUME_DRAIN_MS` against the *late* boundary; the one that mattered was never stated.

### 4. `steamFrozenThisOpen` ignored `suspendSteam`'s return — Low

`steamPid` is cached for the session, so after Steam restarts the SIGSTOP fails (ESRCH/EPERM) while the flag still read "frozen" — every later close then took the deferred-release path and held grabs ~750 ms for a drain that never happens. Now assigned from the return value; the watchdog only starts if the freeze actually took.

### Tests were weaker than the count suggested

Fixtures were `Buffer.alloc(64)` with bits set at the same offsets the implementation reads — they'd pass unchanged if every offset were wrong in the same way. Every fixture emitted `01 00 00 00`; a real frame is `01 00 09 40`. And the "ignores non-input reports" tests forged **byte 0**, baking finding #1 into the suite.

Fixtures now carry a realistic header and forge `ucType` instead. Added direct `isDeckStateReport` coverage. **Mutation-verified:** reverting to the byte-0-only check now fails 5 tests; before this commit it failed none.

### What the review confirmed

Every byte offset in `decodeNavState` was independently corroborated against three sources — `hid-steam.c`, SDL's `SteamDeckButtons`, and InputPlumber's `PackedInputDataReport`. **No offset failed.** Stick Y negation is right (the kernel does `-steam_le16(data + 50)`, matching the evdev path's down-positive convention). The dead-band (0.10) sits safely below both consumers (NavController 0.5, right-stick scroll 0.15).

Also confirmed clean: `shutdown()` not tracking `pendingReleaseTimer` is safe in both orderings (no ioctl on a closed fd), Steam is guaranteed thawed before `process.exit(1)`, the freeze watchdog doesn't interact with the drain window, and SIGKILL-while-open recovery is intact.

### Still outstanding

- **No Deck hardware validation.** Everything Deck-specific remains unverified on real Jupiter/Galileo silicon; that's still where this PR's risk lives.
- The deferred-release timer state machine lives in module-scope `index.ts` and has no test coverage — findings 3 and 4 were both in that untested region.

Verification: `typecheck` clean, `eslint` clean (2 pre-existing warnings in a test file), `test:backend` 2535 pass / 1 fail (the known `sound-loader` flake, also on `main`), `test:ui` 378/378.

srsholmes:
## On-device testing complete — Deck + APEX ✅

Tested the installed build on both host types; behaviour looks good on each.

**Steam Deck (Jupiter, built-in controller)**
- Overlay captures the built-in controller: hidraw nav path active, lizard-mode nodes held grab-only, Steam frozen while the overlay is open in game mode.
- Nav works from BPM home (the case that previously had no input source at all, since the virtual X360 pad only exists while a game is running).

**APEX (IP host)**
- Behaviourally unchanged, as designed — `decideInputStrategy()` keeps IP hosts on the legacy path. No regressions observed against the test guide in the earlier comment.

Together with the green CI runs and the post-review hardening in 27fdbbb / d07a1af (deferred evdev release after resume, watcher restart on death, Guide+B wake restored, Steam/QAM reserved from wake bindings), this is good to merge.

Known/accepted limitations stay as documented on the PR: pre-IIFE open is touch-only, wake-in-debounce nav leak, and Steam reacting live to a non-paddle wake press at open.

PR #229, opened 2026-07-28
fix(overlay): retry transient plugin-bundle fetches

srsholmes:
## Problem

The overlay fires **every** plugin's bundle fetch in parallel ~3 ms into page load. On a handheld that boots straight into gaming mode, Wi-Fi is often still associating at that exact moment — when the interface activates, Chromium's network-change notifier flushes the socket pool and every in-flight request dies with `net::ERR_NETWORK_CHANGED`, surfaced to JS as a bare `TypeError: Failed to fetch`.

Observed live on an OXP APEX (SteamOS), diagnosed via CEF DevTools on `:9222`:

```
15:35:53  overlay webview boots, fires all 25 bundle fetches at +3ms
15:35:54  wlan0: disconnected -> prepare -> config -> connected
15:35:55  wlan0: dhcp4 new lease 192.168.86.44 -> activated
```

The 8 bundles that had already landed were fine. The 17 still in flight — battery-tracker, display-settings, bluetooth, wifi, storage, network-info, rgb-control, quick-links and the rest — all failed, in fetch order. Nothing re-requested them, so those plugins' home widgets and sidebar icons stayed dead **for the entire session**; only a manual overlay restart cleared it.

`importPluginBundle` already dropped the failed promise from `bundleCache`, so a retry *would* have worked — confirmed by re-importing the identical bundles seconds later — but nothing ever retried.

## Fix

**1. Retry transient bundle fetches** on a `150 / 400 / 1000 / 2500` ms ladder.

| Outcome | Retried? | Why |
|---|---|---|
| thrown `fetch` (network layer) | yes | `ERR_NETWORK_CHANGED`, conn reset |
| 5xx, 408, 429 | yes | server hiccup / transient |
| other 4xx | **no** | a real answer: plugin gone (404) or unauthorized (401) |
| module `import()` failure | **no** | a build error; re-importing fails identically |

Cache-busts **per attempt**, so a retry is never handed the failed response out of the HTTP cache. The ladder is a ~4 s wall-clock budget, not "4 chances" — a flushed request fails instantly, so each attempt is exposed for only a loopback round trip. The later rungs exist because `NetworkChangeNotifierLinux` fires per netlink `RTM_NEWADDR` (v4 lease, v6 link-local, v6 SLAAC) and the affected host brings up both eth0 and wlan0.

**2. `usePluginIcons` no longer negative-caches failures.** `iconCache` has no invalidation and its guards short-circuit on `.has()`, so a failed icon load burned that plugin to the letter fallback for the whole page lifetime — and the overlay window is shown/hidden, never reloaded, so that meant the whole session. This was the same bug, unfixed by the retry. Only the success path caches now, where `null` legitimately means "exports no icon".

**3. `runStartupInits` had no retry at all** (`pluginInit.ts`). It fetched `/api/plugins` with a single attempt in the same page-load window, so the same flush killed it — and when it did, **no `loadOnStartup` plugin ever initialised**, silently, for the session. One dropped request took out every startup plugin.

**4. Four hand-rolled retry loops consolidated** into `lib/retry.ts`, covering both shapes the overlay needs:
- `fixedDelays` — bounded ladder, then surface the error (bundles, plugin list).
- `exponentialDelays` — unbounded, for startup gates with no failure path (session token, sidebar list).

`HttpError` carries the status, so retriability is a structural check rather than sniffing an error message. `usePlugins` keeps its effect-scoped cancellation via `shouldContinue`.

## Verification

**34 tests** — 22 for `retry.ts` (both policy shapes, cancellation, error-identity preservation, cache-buster `?`-vs-`&`), 12 for the bundle path. `typecheck` and `eslint` clean. `test:ui` 378/378; `test:backend` 2514 pass / 1 fail (a pre-existing `sound-loader` flake on `main`, untouched here). Full `bun run build` succeeds, so bundling is confirmed, not just types.

Mutation-tested rather than assumed. Eight mutations, each caught by the intended test: `fixedDelays` off-by-one, losing the original error on give-up, 404-treated-as-retriable, hoisting the `shouldContinue` check out of the loop, and four ladder variations. The `shouldContinue` one initially **hung** the suite instead of failing it, so that test was rewritten onto a bounded ladder to fail cleanly.

Note the 12 bundle tests pass unchanged across the `retry.ts` extraction, which is the evidence the refactor is behaviour-preserving — but they were added by this PR's own first commit, not inherited from `main`.

## Known gaps

- **Not exercised on-device.** Reproducing the original bug needs a boot-time network race; the test hardware is currently running another branch. Worth one cold boot before release.
- **`pluginInit` and `usePlugins` have no tests.** Replacing the startup ladder with `[]` — deleting the retry item 3 is named after — passes the suite. So does dropping `headers` from `fetchTextNoCache`, which would strip `Authorization` and 401 the plugin list. The 22 new tests all land on `retry.ts`, already the best-covered part.
- **`initBackend`'s "never rejects" is now configurational, not structural.** It relies on `isRetriable: () => true`; deleting that line passes the suite, and both boot callsites do `initBackend().then(…)` with no `.catch()`, so a rejection would take out the whole boot chain including the `__overlay/show` subscription. The old `for(;;) { try{}catch{} }` made that impossible by construction.
- **`exponentialDelays({ base: 0 })`** yields `NaN` at attempt ≥1024 and would spin. Unreachable from either production callsite (`base: 1000`), but the helper is exported.
- `fetchTextNoCache` appends the cache-buster with string concatenation, so a URL with a fragment would put it in the fragment. No current callsite produces one.

### DIFF
--- a/apps/loadout-overlay/src/overlay/hooks/usePluginIcons.ts
+++ b/apps/loadout-overlay/src/overlay/hooks/usePluginIcons.ts
@@ -38,8 +38,14 @@ async function loadIcon(pluginId: string): Promise<PluginIconComponent | null> {
       iconCache.set(pluginId, icon);
       return icon;
     } catch (err) {
+      // Deliberately NOT cached. `iconCache` has no invalidation, so caching
+      // a failure here burns the icon to the letter fallback for the whole
+      // page lifetime — and the overlay window is shown/hidden, never
+      // reloaded, so that means the whole session. A transient fetch failure
+      // (see fetchBundleSource: a network-change socket-pool flush during
+      // boot) must stay retryable on the next sidebar mount. Only the success
+      // path caches, where `null` legitimately means "plugin exports no icon".
       console.warn(`[usePluginIcons] Failed to load icon for ${pluginId}:`, err);
-      iconCache.set(pluginId, null);
       return null;
     } finally {
       inFlight.delete(pluginId);
--- a/apps/loadout-overlay/src/overlay/hooks/usePlugins.ts
+++ b/apps/loadout-overlay/src/overlay/hooks/usePlugins.ts
@@ -1,5 +1,6 @@
 import { useState, useEffect, useCallback } from "react";
 import { authHeaders, apiUrl } from "../lib/backend";
+import { exponentialDelays, retryAsync } from "../lib/retry";
 import { onConnect, subscribe } from "@loadout/ui/ws-client";
 
 export interface PluginInfo {
@@ -63,18 +64,24 @@ function usePluginList(path: string) {
 
   useEffect(() => {
     let cancelled = false;
-    let retryTimer: ReturnType<typeof setTimeout>;
 
-    async function fetchWithRetry(attempt = 0) {
-      if (cancelled) return;
-      const ok = await fetchPlugins();
-      if (!ok && !cancelled) {
-        const delay = Math.min(1000 * 2 ** attempt, 10000);
-        retryTimer = setTimeout(() => fetchWithRetry(attempt + 1), delay);
-      }
-    }
-
-    fetchWithRetry();
+    // Unbounded backoff: the sidebar is useless without a plugin list, so
+    // there's no failure path worth surfacing — keep waiting for the server.
+    // `shouldContinue` is what makes this safe in an effect: it's checked
+    // before every attempt AND after every wait, so an unmount stops the
+    // chain without firing another request.
+    void retryAsync({
+      label: "[overlay] plugin list",
+      delays: exponentialDelays({ base: 1000, cap: 10000 }),
+      isRetriable: () => true,
+      shouldContinue: () => !cancelled,
+      run: async () => {
+        if (!(await fetchPlugins())) throw new Error("plugin list fetch failed");
+      },
+    }).catch(() => {
+      // Only reachable via cancellation (unmount) — fetchPlugins already
+      // logged anything worth seeing.
+    });
 
     // Re-fetch whenever the WebSocket (re)connects — server is up
     const unsub = onConnect(() => {
@@ -93,7 +100,6 @@ function usePluginList(path: string) {
 
     return () => {
       cancelled = true;
-      clearTimeout(retryTimer);
       unsub();
       unsubChanged();
     };
--- a/apps/loadout-overlay/src/overlay/lib/backend.test.ts
+++ b/apps/loadout-overlay/src/overlay/lib/backend.test.ts
@@ -25,7 +25,13 @@ let calls: string[] = [];
 let origFetch: unknown;
 
 /** Queue a scripted sequence of outcomes, one per fetch attempt. Errors are
- *  thrown (network layer); Responses are returned (HTTP layer). */
+ *  thrown (network layer); Responses are returned (HTTP layer). The last
+ *  entry repeats forever, so "fails and keeps failing" is one entry.
+ *
+ *  Footgun: repeating an entry hands back the SAME Response object, so a
+ *  script whose repeated entry is an `ok()` would throw "Body already used"
+ *  on the second read. Only repeat entries whose body is never read (errors,
+ *  or non-2xx, which short-circuit before `.text()`). */
 function scriptFetch(results: FetchResult[]): void {
   let i = 0;
   (globalThis as { fetch: unknown }).fetch = (url: string) => {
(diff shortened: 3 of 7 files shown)


srsholmes:
## Review response — two independent passes, five fixes applied (`ad72b5b`)

One lens on retry-logic correctness, one on runtime behaviour and codebase fit. Both converged on the same top finding, which was a real bug this PR claimed to fix but didn't.

### 1. `usePluginIcons` negative-cached failures permanently — **the PR's stated residual was wrong**

`usePluginIcons.ts:42` did `iconCache.set(pluginId, null)` in the catch. `iconCache` is module scope with no invalidation anywhere, and the guards at `:29`/`:64` short-circuit on `.has()` — so a failed icon load burned that plugin to the letter fallback for the whole page lifetime. The overlay window is shown/hidden, never reloaded, so that's the entire session.

That is precisely the symptom this PR exists to kill, and it survived the retry ladder. `bundleCache` correctly deletes on failure so widgets get a genuine second chance on remount; icons did not. I described this as "latches until the component remounts" — for icons, remounting never healed it.

The catch no longer caches. Only the success path does, where `null` legitimately means "plugin exports no icon".

### 2. `?? 1000` was dead code masking an off-by-one

Reviewer 1 mutated `delays[attempt - 1]` → `delays[attempt]` and **all 9 original tests still passed** — the ladder silently became 400/1000/1000 because the out-of-range read was swallowed by the fallback. Removed; the index is provably in range.

### 3. `startsWith("HTTP ")` replaced with structure

Neither reviewer found a live false positive (the nearest candidate, `HTTP/2 stream closed`, doesn't match — the guard needs the trailing space). But correctness shouldn't rest on a message prefix. The 4xx throw now sits outside any `try` that could catch it, which also makes the body-read retry explicit rather than incidental.

### 4. Fourth rung, and a corrected rationale

`NetworkChangeNotifierLinux` fires per netlink `RTM_NEWADDR` — v4 lease, v6 link-local, v6 SLAAC — and the affected host brings up **both** eth0 and wlan0, which can spread past the old 1.55 s budget. Ladder is now `[150, 400, 1000, 2500]`; nothing blocks on the outcome, so it's free.

I also corrected the ladder's justification. It claimed the short ladder was safe because `initBackend` proves the server is reachable — but the server is on loopback and answers with no network at all, so its reachability says nothing about whether an interface is about to flush the socket pool. Non-sequitur, now fixed in the comment.

### 5. Test gaps — both surviving mutations now fail the suite

- Backoff test used a one-rung ladder with a lower bound only, so it couldn't detect a wrong rung index. Now two rungs (`[40, 250]`), both bounds asserted.
- Added 5xx-ladder-exhaustion: deleting `lastErr = err` on that branch left the old suite green while degrading the rejection to `…: undefined`, losing the status from the only diagnostic a user sees.
- Added mid-stream body failure and empty-ladder cases.

Verified by re-running both mutations: each now fails exactly one test. 12 tests, 31 assertions.

### Confirmed non-issues (verified, not assumed)

- **Thundering herd / jitter** — Chromium caps 6 sockets per host group, and the requests fail in waves as sockets free, so retries are already staggered. Jitter would be cargo-cult.
- **Startup latency** — `attempt > 0` guards the sleep; attempt 0 is byte-identical to the pre-PR path. Zero cost when everything succeeds.
- **`bundleCache` race** — `promise.catch()` is registered before `bundleCache.set()`, so a delete can never remove a newer entry. Retries strictly improve it by de-duplicating concurrent callers onto one pending promise.
- **Premise** — `CloseAllConnections(ERR_NETWORK_CHANGED)` does not exempt loopback and also fails *pending* (not-yet-socketed) requests, which is why 17 died rather than just the ~6 with sockets.

### Follow-ups, deliberately not in this PR

- `runStartupInits` (`pluginInit.ts:46-55`) fetches the plugin list with a single attempt and would be killed by the same flush — if it is, no bundle fetch is issued at all. `usePlugins.ts` already has backoff for its equivalent call.
- Three hand-rolled retry loops now exist in the overlay (`initBackend`, `usePlugins.fetchWithRetry`, `fetchBundleSource`). A shared `retryFetch({ delays, isRetriable })` is warranted; both reviewers called it a follow-up rather than a blocker.

`bun run typecheck` clean, `eslint` clean, `test:ui` 378/378. `test:backend` 2492 pass / 1 fail — the pre-existing `sound-loader` flake, untouched by this PR.

Still not exercised on-device: the test hardware is running a different branch's overlay build.

Issue #230, opened 2026-07-28
custom TDP values should be respected in the power profiles


srsholmes:
If a user sets a custom TDP range, the power profiles should be updated to reflect their custom values. Maybe we add in to the form an additional optional collapsible region which allows them to create their own power profiles based on their custom values. If left blank make an educated guess based on a deterministic formula.

Also there was reported a bug that the custom tdp profile limits is not respected and the limit is still there on the slider.

Eg. A max wattage of 80w clamps to the default 30w on an unknown device.

PR #231, opened 2026-07-28
fix(tdp-control): respect custom TDP range in power profiles (#230)

srsholmes:
Custom devices now derive their battery cap and power presets from the
TDP range, so a user only has to enter a name plus Min/Max. The presets
and battery cap move into an optional collapsible region in the
custom-device form; anything left blank is filled by a deterministic
formula (Silent=min, Balanced=midpoint, Performance=max, battery cap=max).

This fixes the reported bug where a raised Max TDP (e.g. 80W) still
clamped to the low auto-detected battery cap (~30W) on an unknown device:
the form no longer seeds the battery cap from the detected value, and a
blank battery cap now defaults to the full max instead of throttling.

Also raise the per-game profile engine's sanity clamp from 3-80W to
1-200W (mirroring the custom-device bounds) so custom devices above 80W
aren't silently limited through the per-game path. The real per-device
clamp still happens in the backend's applyTdp().

Claude-Session: https://claude.ai/code/session_01JQ3SbER5nvqs1LQXwWt48f

### DIFF
--- a/plugins/tdp-control/app.spec.tsx
+++ b/plugins/tdp-control/app.spec.tsx
@@ -141,6 +141,45 @@ describe("tdp-control plugin", () => {
     });
   });
 
+  it("keeps presets & battery cap in an optional collapsed region", async () => {
+    const container = document.createElement("div");
+    const { mount } = await import("./app");
+    mount(container);
+
+    await waitFor(() => {
+      expect(container.textContent).toContain("15W");
+    });
+
+    // Open the custom-device form.
+    const gear = container.querySelector(
+      '[aria-label="Custom device settings"]',
+    ) as HTMLButtonElement;
+    fireEvent.click(gear);
+
+    await waitFor(() => {
+      expect(container.textContent).toContain("CUSTOM DEVICE");
+    });
+
+    // Required fields are always visible; the presets/battery cap live in a
+    // collapsed "optional" region and aren't rendered until it's expanded.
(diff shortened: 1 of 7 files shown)


srsholmes:
Closing in favour of a narrower PR off `main`.

The investigation confirmed there **was** a real bug (two, in fact) — see below — but they're fixable without the form redesign, and the redesign isn't wanted:

**Bug 1 — the battery cap silently overrides Max TDP.** Reproduced against `main`: `validateCustomDevice` accepts `maxTdp: 80` with `batteryMaxTdp: 30`, and `effectiveMaxWatts()` (`backend.ts:1302`) returns the battery value whenever `acPowerOnline === false`, which the slider binds to directly. The old form seeds the battery field from the *auto-detected* device, so raising Max TDP without also editing that field leaves the low cap in place — with nothing in the UI explaining why. It only manifests on battery, which is why it reads as intermittent.

**Bug 2 — per-game profiles hard-clamped at 80 W.** `clampTdp` in `tdp-profiles.ts` used a device-independent `MAX_TDP_WATTS = 80`. This silently caps OneXPlayer Super X (90 W) and GPD Win 5 (85 W) — shipped devices, no custom device involved. This PR's 1–200 widening was correct and is being carried over.

The replacement PR keeps the existing form and adds a hard 55 W battery safety ceiling with a warning banner, which also makes bug 1 impossible to hit by accident. The derived-presets idea from #230 is left open as a separate question.

PR #233, opened 2026-07-28
fix(tdp-control): 55W battery ceiling; stop the battery cap silently overriding Max TDP

### LINKED ISSUE #230: custom TDP values should be respected in the power profiles

srsholmes:
Closes #230. Replaces #231 (closed) — same bugs, without the form redesign.

## The two bugs, both reproduced against `main`

**1. The battery cap silently overrode Max TDP.**

```
validation accepts max=80 with batteryMax=30 : true
  acPowerOnline=true  -> slider max = 80W
  acPowerOnline=false -> slider max = 30W    ← the reported symptom
  acPowerOnline=null  -> slider max = 80W
```

The form seeded **Battery max TDP** from the *auto-detected* device via `fillFromDetected()`. Raise Max TDP to 80, leave that field at a detected 28–30 W, and `effectiveMaxWatts()` returns the low figure whenever `acPowerOnline === false` — which the slider binds to directly. Validation accepts it, because 30 genuinely is between min and max.

Two things made it read as a bug rather than a mis-filled field: it **only bites on battery** (plugged in, the slider goes to 80 as expected), and **nothing in the UI said why**.

**2. Per-game profiles were hard-clamped at 80 W.**

```
OneXPlayer Super X   device max  90W -> per-game profile saves 80W   <-- SILENTLY CLAMPED
GPD Win 5            device max  85W -> per-game profile saves 80W   <-- SILENTLY CLAMPED
```

`clampTdp` used a device-independent `MAX_TDP_WATTS = 80`. Two **shipped** devices can't have per-game profiles at their real ceiling — no custom device involved, and nobody had reported it.

## Fixes

- **Battery max TDP is now optional.** Blank = "same as Max TDP", and it is no longer pre-seeded from another device's profile. An explicit value is still honoured and validated.
- **`clampTdp` widened to a 1–200 W sanity guard.** The real ceiling is applied by `applyTdp()`, which is device- and power-state-aware.
- **New hard 55 W on-battery ceiling for every device** — `BATTERY_SAFE_MAX_WATTS` in `@loadout/devices`, applied in `effectiveMaxWatts()`. Sustained draw above this pushes cell temperature and discharge current past what these packs are specified for.

Because `applyTdp()` already clamps the hardware write to `effectiveMaxWatts()`, the ceiling is enforced on **every** path — slider, per-game profiles, AC transitions — not just the slider's `max` attribute. It also makes bug 1 unreachable by accident: the worst a blank or wrong battery figure can now do on battery is 55 W.

An unknown/null AC state deliberately does **not** restrict — misreading AC as absent would throttle a plugged-in device for no reason.

## UI

A non-dismissable warning banner, mirroring fan-control's safety-override `Alert`:

> **TDP limited on battery** — Capped at 55 W while on battery — sustained draw above this degrades the battery. Plug in to use the full 80 W.

It fires **only when we are actually overriding the user**. A device whose own battery figure is already below 55 W shows nothing, because that figure is their choice, not our override. The `acPowerChanged` event carries the flag so the banner appears on unplug without a refetch.

## Behaviour change worth calling out

**OneXPlayer Super X** ships `batteryMaxTdp: 65` and is now capped to **55 W** on battery. Intentional, per the safety rationale above.

## Verification

`typecheck` clean, `eslint` clean, `test:ui` 378/378, `test:backend` 2576 pass / 1 fail (the pre-existing `sound-loader` flake, also on `main`).

11 new tests: the ceiling never being exceeded regardless of device or override, lower device figures *not* being raised to it, AC-unknown not restricting, banner-activation conditions, and optional/blank/null/explicit battery-max validation — including the exact #230 numbers (max 80 + stale 30 → 30 before, 55 after).

Four existing tests asserted the old 3–80 clamp, i.e. they encoded bug 2. Updated, including one that now proves a 90 W Super X profile round-trips intact.

## Not included

The derived-presets feature from #230 (auto-filling Silent/Balanced/Performance from the range). That was the part of #231 that drove the form redesign; it's a separate product question and #230 stays open for it if wanted.

### DIFF
--- a/packages/devices/src/devices.test.ts
+++ b/packages/devices/src/devices.test.ts
@@ -1,5 +1,11 @@
 import { describe, it, expect } from "bun:test";
-import { matchDevice, matchProfileName } from "./devices";
+import {
+  matchDevice,
+  matchProfileName,
+  effectiveMaxWatts,
+  isBatterySafetyCapActive,
+  BATTERY_SAFE_MAX_WATTS,
+} from "./devices";
 
 describe("matchDevice", () => {
   it("matches a known device by DMI product-name substring", () => {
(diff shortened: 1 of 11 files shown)


srsholmes:
## Banner reworked (`4c09a54`)

Per feedback — it's now an informational notice rather than a warning, and it sits with the control it explains.

| | Before | Now |
|---|---|---|
| Variant | `warning` | `info` |
| Placement | Top of page | Under the slider, inside the CURRENT TDP card |
| Shows when | Only when the global 55 W cap binds | Whenever being on battery lowers the ceiling at all |
| Dismissable | No | Yes |

**Copy:** *"On battery, so the maximum is 25 W. Plug in for up to 30 W."*

### Shows regardless of what the user set

`isBatteryLimited()` replaces `isBatterySafetyCapActive()`. It asks the question that actually matters — *"is my max lower than usual right now?"* — without caring whether the device's own `batteryMaxTdp` or the global cap produced it. Distinguishing them was reviewer-brain, not user-brain.

It stays false when the two ceilings are equal (Steam Deck at 15 W on battery and plugged), because there "the max might be lower" would simply be untrue.

### Dismissal and re-show

`Alert` gained an opt-in `onDismiss`. It remains **non-dismissable by default** — most of its uses (fan-control's safety override, "this won't work in your session") are load-bearing and shouldn't be silenceable — so this is opt-in per callsite, and the caller owns both the hidden state and the policy for when it returns.

Dismissal stores the **token** of the situation dismissed (the battery/plugged ceiling pair), not a boolean. The notice comes back when:

1. **The situation changes** — unplugging, replugging, or editing the device's limits all mint a new token that no prior dismissal covers.
2. **The user returns to the plugin** — local state, and `tdp-control` isn't `keepAlive`, so navigating away clears it.
3. **The slider is dragged up to the ceiling** — pressing against the limit is exactly when the explanation is wanted again.

It deliberately does *not* reappear on every TDP change; dismissing while already at the ceiling keeps it hidden until one of the above actually happens. Nagging would undo the point of making it dismissable.

### Verification

`typecheck` clean, `eslint` clean (3 pre-existing `any` warnings in the spec's mock block, above my diff). `test:ui` **384/384**, `test:backend` **2578 pass / 1 fail** — the pre-existing `sound-loader` flake, confirmed by name and also failing on `main`.

6 new UI tests: copy names both ceilings, `alert-info` and not `alert-warning`, renders *after* the slider in document order (`compareDocumentPosition`), hides on dismiss, absent when the ceilings match, absent on AC.

PR #235, opened 2026-07-29
fix(uninstall): revert the InputPlumber wake-button mapping

### LINKED ISSUE #234: uninstall.sh leaves the InputPlumber wake-button mapping in place (button stops opening the OSK)
## Summary

`scripts/uninstall.sh` removes the services, binary, overlay tree and desktop file, but it never reverts anything the **input-plumber** plugin wrote. On a non-Deck handheld the overlay wake button *is* an InputPlumber profile, so after uninstalling, the user's physical button stays remapped to `KEY_F16` instead of doing what the OS normally does with it (on an OXP Apex, opening the on-screen keyboard).

Reported by a user on an Apex who uninstalled Loadout and lost their keyboard button.

## What gets left behind

Written by the plugin (`plugins/input-plumber/lib/wake-trigger.ts`, `lib/profile.ts`), removed by nothing:

- `/etc/loadout/inputplumber/overlay-profile.yaml` — the `<button> -> KeyF16` mapping, live-loaded onto the composite device via `LoadProfilePath`
- `/etc/udev/rules.d/71-loadout-inputplumber-uaccess.rules`
- `/etc/udev/rules.d/72-loadout-deck-hidraw.rules` (Deck)
- InputPlumber itself, when Loadout installed it via `plugins/input-plumber/scripts/install-inputplumber.sh` (unit, `/var/lib/inputplumber`, dbus policy, polkit files)
- `/etc/inputplumber/devices.d/50-steam_deck.yaml` with `auto_manage: true`, for anyone who installed a version predating #87 — that one keeps IP holding the Deck controller away from Steam Input

`grep -n inputplumber scripts/uninstall.sh` returns nothing.

## Why the button stays broken

The profile only lives in InputPlumber's runtime state, so restarting IP or rebooting *does* clear it — the backend is gone and nothing re-loads it. But nothing in the uninstall flow tells the user that, and the failure mode is silent and confusing: the button just stops opening the keyboard. Same regression shape as #206, minus any log to grep.

## Proposed fix

1. **Clear the binding before teardown.** While the backend is still installed, run the plugin's `clearWakeButton()` equivalent (render `renderClearedProfile` + `LoadProfilePath`) so the controller is handed back in its default state without needing an IP restart.
2. **Remove the files.** `rm -rf /etc/loadout`, drop the two `71-`/`72-loadout-*` udev rules, `udevadm control --reload`.
3. **Restart InputPlumber** as a belt-and-braces step (`systemctl reset-failed inputplumber && systemctl restart inputplumber`) so the change lands without a reboot.
4. **Leave InputPlumber installed.** It's the input layer the handheld fleet depends on and it predates Loadout on most of them. If we want to offer removal, it should be an explicit opt-in prompt gated on us having synthesised the unit ourselves (the `steam-loader install` Description is a reliable marker), never the default.
5. **Deck-only:** offer to remove `/etc/inputplumber/devices.d/50-steam_deck.yaml` if it exists and carries our `auto_manage: true`.

All of the above is sudo-gated, so it should follow the same pattern as the existing system-service removal block and degrade gracefully when sudo isn't available.

## Docs

Add to the README uninstall section: set the wake button to **Off** in the InputPlumber panel before uninstalling, which is the clean path today.

## Workaround for affected users

```sh
sudo rm -rf /etc/loadout
sudo rm -f /etc/udev/rules.d/71-loadout-inputplumber-uaccess.rules
sudo udevadm control --reload
sudo systemctl reset-failed inputplumber && sudo systemctl restart inputplumber
```

A plain reboot also works. `sudo evtest` on the "InputPlumber Keyboard" device confirms the diagnosis — `KEY_F16` on button press means the profile is still loaded.

srsholmes:
Fixes #234.

## Problem

The overlay wake button isn't a Steam setting — the input-plumber plugin writes an InputPlumber profile remapping a physical button to `KEY_F16` and live-loads it over DBus. `uninstall.sh` removed the services, binary, overlay tree and desktop file, but nothing it did touched that profile.

So after uninstalling, the user's button stayed bound to F16 and stopped doing what the OS normally does with it — on an OXP Apex, opening the on-screen keyboard. Silent failure, no logs, nothing pointing back at Loadout.

## What this does

`revert_inputplumber()` runs right after the services are stopped (so the backend can't re-apply the profile between the delete and the restart):

- removes `/etc/loadout/inputplumber`, and `rmdir`s `/etc/loadout` only if it's empty — never a blind `rm -rf` of a directory other plugins may grow into
- removes `71-loadout-inputplumber-uaccess.rules` and `72-loadout-deck-hidraw.rules`, then `udevadm control --reload`
- removes the Deck `auto_manage` override from versions predating #87, **only** when the file carries our `Generated by Loadout` marker — that path is a general-purpose InputPlumber override location a user may have authored themselves
- restarts InputPlumber, since the profile lives in IP's runtime state and deleting the file alone doesn't unbind the button until IP rebuilds its composite devices. `reset-failed` first so a tripped start-limit can't block it, same as the plugin's own restart path.

## What it deliberately doesn't do

**InputPlumber stays installed and enabled**, even where Loadout installed it from scratch via `install-inputplumber.sh`. It's the input layer the handheld fleet depends on and it predates Loadout on most devices — a stock IP with no Loadout profile loaded is exactly the state the device would be in had Loadout never run. Ripping it out would be the genuinely destructive option and would break controllers for people who never had a problem.

The issue also proposed clearing the binding via the plugin's `clearWakeButton()` before teardown. I skipped it: doing that from shell means re-rendering the cleared profile YAML and driving `LoadProfilePath` over `busctl` with the device's live target list, which is a lot of fragile surface to duplicate outside TypeScript. The IP restart reaches the same end state — default mapping, no reboot needed — in three lines.

## Recovery for already-affected users

Every step is guarded on file existence, so **re-running the uninstaller on a machine where Loadout is already gone** cleans up the leftover mapping. That's the fix for anyone who uninstalled with an older version; the README now says so. No need to reinstall Loadout first.

## Failure handling

Each privileged step warns instead of aborting (`|| warn`), so a sudo prompt the user declines or a failed restart still leaves a working system, prints what to do, and lets the rest of the uninstall finish. If the restart fails the user is told to reboot, which reaches the same state.

## Testing

No shell test harness exists in the repo, so I drove the function directly with every system path redirected into a sandbox and `as_root`/`systemctl`/`udevadm` stubbed. Cases covered:

| Scenario | Result |
|---|---|
| Nothing installed | No-op, exit 0, `"No InputPlumber changes to revert."` |
| Full install, IP running | Files removed, IP restarted |
| IP not running | Files removed, no restart attempted |
| Deck override **without** our marker | Preserved |
| Deck override **with** our marker | Removed |
| Only udev rules left (profile already gone) | Rules removed, no pointless restart |
| Restart fails | Warns, exit 0, uninstall continues |
| `/etc/loadout` has a sibling plugin dir | Parent preserved |

`sh -n` clean. Not yet run end-to-end on hardware — worth one pass on the Apex before release, since that's the reported device.

## Follow-up

CHANGELOG isn't touched here — per `docs/releasing.md` the version section is written at release time. This wants a **Fixed** line when v0.7.3 is cut, along the lines of: *uninstalling Loadout now gives your wake button back — it previously stayed bound to the overlay and stopped opening the on-screen keyboard.*

https://claude.ai/code/session_01LYHLBdkAnawV7XkYJXoUE5


srsholmes:
Reviewed; three defects found, all fixed in bf8136f. All three were in the code *around* the new revert rather than in it.

**1. Medium — a failed sudo step skipped the revert entirely.** `sudo rm -f "$SYSTEM_SERVICE_FILE"`, its `daemon-reload`, and the earlier `systemctl --user daemon-reload` had no `|| true` / `|| warn` while every sibling around them did. Under `set -e` any of them killed the script silently, and the revert call sits downstream of all three — so a user who declines or mistypes the sudo password got a half-finished uninstall with the button still bound to F16 and nothing on screen. This made the PR's own graceful-degradation claim false for the steps gating the fix.

Verified end-to-end with `HOME` sandboxed and `sudo`/`systemctl` stubbed to fail:

| | exit | reached revert | reached end |
|---|---|---|---|
| before bf8136f | 1 | no | no |
| after | 0 | yes | yes |

**2. Low — `udevadm control --reload` with no `udevadm trigger`.** The ACLs Loadout granted survived until the next reboot, since the nodes are never re-enumerated. The install side always pairs them (`wake-trigger.ts` `ensureUaccessRule`, `wake-trigger-deck.ts` `ensureDeckHidrawUaccess`), and the two rules need *different* triggers — they're now tracked separately, each getting the trigger its install counterpart uses.

**3. Low — claimed to revert InputPlumber on machines that don't have it.** Post-#87 the Deck path is a native hidraw watcher and SteamOS ships IP disabled or absent, but finding only `72-loadout-deck-hidraw.rules` was enough to print "Reverting the InputPlumber wake-button mapping" and "InputPlumber itself is left installed and enabled". Messaging is now neutral by default and only mentions InputPlumber when an IP artefact is actually present; same correction to the README.

The review also refuted a few things worth recording as checked, not overlooked: the marker grep fails *safe* on an unreadable file (tested at mode 000), `rmdir` cannot touch a non-empty `/etc/loadout`, the hardcoded paths match `profile.ts` exactly with no drift, the marker string matches what pre-#87 actually wrote, and restarting IP is genuinely sufficient — `LoadProfilePath` is the only load path and `reloadPersistedProfile()` is called only from the backend's `onLoad`, so a stopped backend cannot re-apply the profile.

Still unverified: nothing has run on real hardware, and the assumption that InputPlumber doesn't persist a `LoadProfilePath`'d profile across a restart was checked against all repo code but not IP's own source. One pass on the Apex before release is still worth doing.

srsholmes:
Second review round. Four more findings, all fixed in c1f303f. Round 1's fixes for the udev triggers and the messaging both hold; the sudo-guard fix held only upstream of the revert.

**1. Medium — the restart gate keyed off the wrong signal.** It required the profile directory to exist on disk, but a mapping can be live in InputPlumber's runtime with the profile already deleted. That's exactly the state anyone following #234's workaround lands in — it starts with `rm -rf /etc/loadout`, and some users will stop before the restart line. They'd run the uninstaller, get `[OK] Wake-button files removed`, and still have a button bound to F16.

The gate is now `ip_uses_ip` — any IP-related artefact, plus a new signal: a persisted non-Deck binding in `$XDG_CONFIG_HOME/loadout/plugins/input-plumber.json`. That file is written by the root backend but under the user's `HOME` (`loadout.service` sets `Environment=HOME=__HOME__`), so the uninstaller can read it. It also covers the case where every file is already gone but the mapping is still loaded, where the script previously said "nothing to revert" and did nothing. `deck:<button>` bindings are excluded — they never touch IP. An unnecessary restart costs a one-second controller blink, which is the right side to err on.

**2. Medium — `sudo rm -f "$BINARY_PATH"` was still bare.** The defect bf8136f fixed upstream of the revert survived downstream of it: on every non-SteamOS host a declined sudo password aborted the script there, skipping the plugin-data and configuration prompts, leaving the overlay tree on disk, and never printing the final banner. bf8136f's commit message claimed those three commands were the only unguarded ones — that was wrong. This is the fourth, and per a line-by-line pass over the rest of `main()` it's the last; everything else targets a user-owned path.

**3. Low — false success over failed removals.** `[OK] Wake-button files removed` printed even when every `rm` had failed. Now tracked, and leftovers are reported explicitly.

**4. Low — README overpromised.** "Re-run and it will clean up the leftover mapping" is false if the binding record is gone too, so the reboot fallback is now stated.

`ip_had_mapping` is removed rather than left dead.

Testing: 9 reachable artefact states (profile / `71-` rule / `72-` rule / marked Deck override / persisted binding × IP up-or-down), asserting both whether IP restarts and whether the messaging mentions InputPlumber at all, plus the all-sudo-denied path end-to-end through `main()`. `sh -n` and `bash --posix -n` clean.

One finding I did **not** act on — worth your call. Round 2 noted that bf8136f's `|| true` on `systemctl --user daemon-reload`, while correctly stopping the abort, newly opens a path where the script proceeds to `rm -rf` the overlay tree while a live overlay process may still hold an `EVIOCGRAB` on an external pad (SSH without a pty into a machine with an active session: `systemctl --user is-active` fails, so the stop branch is skipped entirely). The gap is pre-existing and the fix — a `pkill` fallback — adds process-killing behaviour rather than correcting a claim, so I'd rather you decide than have me slip it into a bugfix PR.

Still unverified, unchanged: nothing has run on real hardware, and whether InputPlumber itself persists a `LoadProfilePath`'d profile across a daemon restart is inferred from this repo's code and #234's own description, not read from IP's source. If it *did* persist, the restart approach would need replacing with a `renderClearedProfile` load.

srsholmes:
Two follow-ups in 70c3d49 — the last open finding, and the assumption both review rounds left unverified.

**The surviving-overlay case is now handled.** When there's no user bus (SSH without a pty, `sudo sh`), `systemctl --user is-active` fails, the stop branch is skipped, and the overlay keeps running — holding its exclusive `EVIOCGRAB` on external pads, which then look dead in Steam until logout, while the script deletes the overlay tree from under the live process. Falls back to TERM → one-second grace to release the grab → KILL. The match pattern is the absolute install directory, so nothing else can match it.

Verified with a fake long-lived process under a sandbox install dir (killed) alongside an unrelated `sh -c` loop (untouched), and confirmed a silent no-op when no overlay is running. I'd flagged this as your call because it adds process-killing behaviour rather than correcting a claim — going ahead on the "do the fixes you think best" steer, so revert this commit alone if you'd rather it landed separately from the #234 fix.

**InputPlumber does not persist a `LoadProfilePath`'d profile across a restart — confirmed from IP's source, not inferred.** `load_profile_path` (`src/dbus/interface/composite_device.rs:153`) → `DeviceProfile::from_yaml_file` → `load_device_profile()` (`src/input/composite_device/mod.rs:436-465`), which mutates in-memory state and emits a DBus property update. There is no `fs::write`, `File::create` or `write_all` anywhere in that module, and IP never writes a profile reference into a device config. So `restart inputplumber` genuinely drops the mapping, and this PR's restart-based approach is correct rather than merely consistent with our own code.

That closes the escalation path round 2 raised: if IP *had* persisted the profile, the restart would have been insufficient and this would have needed a `renderClearedProfile` load instead. It doesn't, so it doesn't.

Remaining known gap is unchanged and now the only one: **none of this has run on real hardware.** Everything is sandboxed execution with `sudo`/`systemctl`/`udevadm`/`as_root` stubbed. One pass on the Apex before release is still worth doing — that's the reported device, and it's the one path (profile present, IP active, restart taken) where a wrong assumption would be user-visible.

srsholmes:
**Verified on hardware — the last open gap is closed.**

Run on a real OneXPlayer Apex (InputPlumber-managed, wake button bound to `Gamepad:Button:Keyboard`), from this branch, with Loadout installed. The keyboard button went back to opening the on-screen keyboard. Post-run state on the device:

- `/etc/loadout` — gone
- `/etc/udev/rules.d/*loadout*` — none left
- `inputplumber` — still **active**, restart timestamp matching the run (so the revert landed in IP's running state, no reboot needed)
- `~/.config/loadout/plugins/input-plumber.json` — retained, binding intact
- `~/.local/share/loadout` — retained

Both prompts were answered "no", so this also exercises the keep-my-data path: the persisted binding survives, which is correct — it's the user's setting, and a reinstall restores it via the backend's `onLoad` reconcile.

That covers the one path I called out as user-visible if an assumption were wrong: profile present, IP active, restart branch taken. Combined with the source-level confirmation that IP doesn't persist a `LoadProfilePath`'d profile, the mechanism is now verified end to end rather than inferred.

Ready to merge from my side. Still outstanding, both release-time: the CHANGELOG line for v0.7.3 (per `docs/releasing.md`), and a decision on whether 70c3d49's overlay-process fallback should land here or separately.

PR #236, opened 2026-07-30
feat(collections): a TabMaster successor — rule engine, Steam collection sync

srsholmes:
Custom, filter-driven library tabs — a successor to the Decky plugin [TabMaster](https://github.com/Tormak9970/TabMaster). You build tabs in Loadout's overlay from a rule tree; each one is kept in step with a real Steam collection, so it shows up in Steam's own library too.

## It was going to be tabs. It isn't, and that's the point

The plugin was called `library-tabs` and aimed at what TabMaster does: extra tabs in Steam's library tab bar. That was built, measured, and dropped. This is worth reading before reviewing, because it explains the shape of everything below.

Steam's tab bar is not data-driven. The array is built inline in a `useMemo` and returned as `rgTabs` — `AllGames`, `Installed`, `Favorites`, and a *single* `Collections` entry — with each tab conditionally pushed. Nothing reads a store. So no amount of correctly written data adds a tab; the only way in is to patch that module, which is exactly why TabMaster patches.

It was implemented, via source patching rather than TabMaster's render-time dispatcher swap, and verified applying on-device: module `97220` patched, runtime loaded with 14 evaluated tabs. The tabs still never rendered — most likely webpack's module cache, since that module had already executed, but that was never proven.

What is not in doubt is the cost. Three separate injector defects sat between a correct patch and a rendered tab, none visible until you were deep in it, and clearing them still left an unverified boot ordering and a module-cache problem. That is a permanent maintenance tax, re-earned on every Steam client update, for a feature Valve can remove without warning.

Collections is the sturdier half and it already works: Steam's own data model, survives Steam updates, survives the plugin being uninstalled, and cannot take the library page down. So the plugin was renamed to `collections` — named for what it delivers rather than what it attempted.

The three injector fixes are **kept**, because they are real defects that would bite the next plugin to attempt a patch:

- the patcher skipped any module over 200,000 chars *before* consulting the find pattern. Steam's library module is 646,312 — so the one module a UI patch cares about was silently unpatchable.
- `$self` expanded via dot notation, so `globalThis.__LOADOUT_PLUGIN_input-plumber` parsed as a subtraction. Every hyphenated plugin id — nearly all of them — got a patch that compiled and then threw.
- the install guard made re-injection a no-op, so restarting the daemon while Steam ran kept whatever patch set the page booted with.

## Why bother at all

TabMaster is the most-loved library-organisation plugin on Steam Deck (~293k downloads) and has no direct competitor. Its job-to-be-done isn't customisation, it's *backlog suppression* — users hide every tab except Installed and Favourites "so I cannot see everything I can install".

It is also structurally fragile in ways this design avoids:

1. **It patches Steam's library route.** When that throws, Steam's library becomes unreachable — users report rebooting ~19 times, locked out of Decky itself and unable to uninstall the plugin that broke them. Nothing here renders inside Steam.
2. **Its data layer is synchronous.** `FilterFunction` is `(params, app) => boolean` with nowhere to `await`; its maintainers cite this to refuse the entire backlog — player counts, ProtonDB, HLTB, friends-playing-now.
3. **Its builder silently produces empty tabs.** No match count until you save.
4. **Config loss is unrecoverable**, per its own maintainer.

## The ideas that carry the design

**Three-valued logic.** A rule returns `true | false | "indeterminate"`, where indeterminate means *the data source couldn't answer*. Combination follows Kleene's strong logic; inversion preserves indeterminate; `indeterminatePolicy` defaults to `"pass"`, so an outage degrades a tab instead of emptying it.

**Async facts, synchronous evaluation.** Anything needing I/O is batch-resolved outside the evaluator and materialised onto `EvalGame.facts` before a sync pass — the structural fix for TabMaster's refused backlog.

**The trace is the product.** One traced evaluation yields per-rule match counts, both combinator outcomes (the `ALL → 0 · ANY → 340` line that makes the AND/OR mistake unmakeable), leave-one-out counts (so an empty tab can say *"remove `Release date before 2015` and 214 games match"*), and per-leaf bitmasks, so contradictory rule pairs are found empirically rather than from a hardcoded table.

**Evaluation runs in the webview.** `app.spec.tsx` asserts that switching tabs issues *no* RPC. That is what makes per-keystroke live counts affordable, and it is precisely what TabMaster's synchronous data layer cannot do.

## The mirror

Tabs are kept in real Steam collections. The planner is pure and CDP-free, because it is the part that can destroy a collection someone spent an evening curating — so it is exhaustively testable with no Steam running, and the UI can show a dry run before anything is written.

Identity comes from a ledger, never from a collection's name. Matching on names is how an automated mirror eats a collection you built by hand and called "Backlog". Each of these is a test:

- a collection whose id we did not record is never deleted
- a name taken by a collection we did not create is reported as a conflict, not adopted
- dynamic and non-editable collections are refused
- drift compares against what we last wrote, not what the tab now matches, so a rule edit doesn't read as user interference
- a failed write is never recorded in the ledger — recording one makes the next run see phantom drift and "fix" it by removing real games
- deletes and renames run before creates, so a freed name is reusable within one sync
- sync twice with nothing changed and the second run writes nothing

Settings carries one switch, "Sync with Steam". Turn it on and every later edit syncs itself; the dry run stays reachable, because it is the only way to see a deletion coming.

## The data layer was the real bug

The plugin originally sourced games from `appmanifest_*.acf` + `shortcuts.vdf` — **installed only**. Of 32 rule kinds, 6 worked, 22 had no data behind them, and 2 were degenerate-always-true. `recent` and `never-played` were structurally always empty. 545 unit tests passed and caught none of it, because they ran against a 10-record fixture where every field was populated.

Reading Steam's own `appTypeCollectionMap` and the `appStore` prototype getters took the library from ~2000 to **4358** games. ROMs are classified from Steam ROM Manager's `srm-` collections, so they can be excluded rather than clogging every tab.

## Testing

`bun run test` → **3209 pass**. One failure, `sound-loader › serializes concurrent apply calls`, is pre-existing and verified against a stashed clean tree. `typecheck`, `lint`, `check:specs` and `check:dead-code` are green.

Three tiers, because unit tests against a fixture are what hid the data bug in the first place:

- **CI, no hardware.** Includes `mirror-agreement.test.ts`, which runs every built-in tab and every template through evaluate → plan → apply → read back against an in-memory Steam, asserting the collection equals `evaluateTab().matched`. That is the only automated guard against a tab meaning one thing on screen and another in Steam.
- **Corpus** (`*.corpus.spec.ts`), against a captured real library. Asserts on the *trace*, not the match list: a rule with `indeterminate === N` matches everything under the `"pass"` policy while looking fine. Carries a ratchet of known-broken rules with expiry dates, so entries can only shrink. The corpus is gitignored — it is a dump of a real library. `docs/library-corpus-profile.md` records aggregate population stats only.
- **Live** (`*.live.spec.ts`), against a running Steam. Contract tests that fail *by name* when Steam moves, plus a write-gated round trip behind `LOADOUT_LIVE_WRITES=1` that asserts against `GetCollectionListForAppID` — the index Steam's library actually renders from — and diffs the entire collection list afterwards to prove nothing else moved.

The new patcher tests *execute* the generated script against a fake `webpackChunksteamui` rather than asserting on its text. Asserting on text is how a patcher passes its suite while silently patching nothing, which is the exact failure that prompted them.

## What four review passes found

The branch was reviewed by four independent passes — rule engine, the Steam write path, UI/config, and a probe of Steam's own JS. Everything below was confirmed by execution and is fixed on this branch, each with a test verified to fail without its fix.

**Four ways to lose a Steam collection.** `setConfig` accepted a client-supplied `mirror.ledger` — backend-owned state — and the UI posts config from React state captured before the last sync's broadcast, so toggling the Settings switch mid-sync reverted the ledger while the collections still existed. Since deletion *requires* a ledger entry, that collection became permanently unreachable while blocking its own tab by name. Nothing serialised syncs, so two of the three entry points overlapping each created a collection for the same tab. An owed sync was silently cancelled when Steam died after the CDP session connected, because each collection then fails individually rather than throwing. And two error paths interpolated a raw `collectionId` into the in-page program while every other interpolation was `JSON.stringify`'d.

**The rule builder was unusable with a gamepad.** The shell drives focus through one `SpatialNavigation` singleton, and a node joins it only by registering through `useFocusable`. "Add rule", both ⋮ menus, every menu action, the chips backing all options/choice/picker rules, and the range and number fields were plain elements — invisible to the D-pad. Every spec passed, because `fireEvent.click` works perfectly well on a button no user can reach.

**A sparse fact map made every unanswered game match.** `rules.ts` is explicit that "the source answered but has nothing for this game" is a definite `false` while "the source is down" is `indeterminate` — that distinction is the whole design. But `buildEvalGames` only wrote a fact when the resolver returned one, so a resolver emitting only what it found left the rest `indeterminate`, which the default `"pass"` policy counts as a match. A ProtonDB rule matched 9 of 10 games with `blockedFacts` empty, so the diagnostics said nothing. `FactValue` gains a `missing` state for exactly this case.

Also fixed: a corrupt settings file was silently replaced by defaults with no backup and no warning; duplicating a group gave every copied rule the same id; a search matching nothing rendered a blank page; and five Tailwind utilities used only by this plugin became inline styles, since plugins ship as a separate release asset from the overlay binary and a class no other source uses has no rule in an older shell's CSS.

## Notable decisions

- **`__core:game-metadata` as a new service**, not an extension of `GameInfo`. `getGames()` is called on mount by ~9 plugins that want a name and a capsule URL; +25 fields makes them all pay.
- **Missing metadata for one game is `false`, not `indeterminate`.** "Steam never recorded this playtime" is not "the HLTB plugin is disabled". The tell was that nearly every builtin needed a policy override to behave sanely.
- **`packages/steam-cdp` gains `collections.ts`** rather than changing `addAppToCollection`, which resolves by name and can only add — right for recomp, wrong for a mirror, and depended on by `protondb-badges`.
- **No drag-to-reorder**, deliberately. Move left / Move right cover it and always will; drag is unusable with a gamepad.


srsholmes:
Superseded by #239, which rebuilds this as a collections manager.

The tab metaphor and the thing it produced were fighting each other: the UI said "Tab options" and "Built-in tabs can be hidden but not deleted", while what actually reached Steam was a collection. It also never read the collections you already have — EmuDeck's `srm-*` sets were invisible to it.

Most of the work here carried over. The tab-specific code was ~1,250 lines of 23,000; the rule engine, mirror planner, windowing, diagnostics and the `steam-cdp` collection APIs all moved across unchanged.

Two pieces were split out rather than dropped:

- **#238** — the webpack-patcher and sound-loader fixes, which never depended on this plugin.
- **#239** — the plugin itself, verified on device listing all 15 real collections.

Closing rather than reworking in place: refactoring this branch was attempted and the type-error count *rose* under sweeping edits, in the two files where a wrong identifier deletes real Steam collections.

PR #253, opened 2026-08-16
feat(tdp-control): prefer firmware-attributes rails over ryzenadj on AMD [parked — needs ROG Ally / Legion Go]

srsholmes:
> **Parked — draft, not intended to merge as-is.** Needs a ROG Ally or a
> Legion Go to validate, and we have no tester on either. Everything in
> this branch that could be shipped without that hardware has been split
> out and merged or opened separately:
>
> - #254 — Wi-Fi 7 `iwlmld` recovery
> - #255 — Intel `i915`/`xe` GPU temperature sensors
> - #256 — Intel handheld TDP + GPU control (AMD paths untouched)
>
> What remains here is the one change that alters behaviour on AMD
> handhelds we can't test.

## What's left in this PR

Making generic `firmware-attributes` discovery outrank `ryzenadj` on
**every** AMD device, rather than only on non-AMD as #256 ships it.

Two coupled effects:

1. **Precedence.** ROG Ally and Legion Go stop using their DMI-matched
   paths and get generically discovered rails instead. Other AMD
   handhelds that expose `firmware-attributes` — msi-wmi-platform, and
   whatever ships next — start preferring those rails over ryzenadj.

2. **Write unit.** Discovered rails infer `scale` from `max_value`
   (= 1, i.e. watts, on asus-armoury and lenovo-wmi-other-N) instead of
   assuming milliwatts. On those devices that is a 1000× change in what
   gets written.

## Why it's probably right, and still parked

Effect 2 looks like a genuine bug fix rather than a regression: `main`
writes `watts * 1000` to `asus-armoury` rails whose `current_value` takes
watts, so a modern-kernel ROG Ally appears to be **broken today** — the
write is rejected and the old read-back path would have reported 0 W.

But "appears" is doing real work in that sentence. Confirming it needs
the hardware. And #256 already carries a cheaper safety net for the same
failure: the DMI-matched paths retry in the other unit when a write is
rejected, which rescues that device without any precedence change.

So the value left in this PR is the generic path for *future* AMD
handhelds, not a fix for a known-broken one. That isn't worth changing
detection order on shipped hardware, unverified.

## To pick this up later

Wanted: a ROG Ally or Legion Go on a kernel with `asus-armoury` /
`lenovo-wmi-other-N` bound. Useful evidence, in order:

1. `cat /sys/class/firmware-attributes/*/attributes/ppt_pl1_spl/max_value`
   — settles the unit question outright.
2. Whether the TDP slider works at all on `main` today.
3. Whether #256's retry-in-the-other-unit path recovers it.

If (1) says watts and (3) works, this PR can probably be closed rather
than merged.

### DIFF
--- a/packages/devices/src/devices.test.ts
+++ b/packages/devices/src/devices.test.ts
@@ -4,6 +4,7 @@ import {
   matchProfileName,
   effectiveMaxWatts,
   isBatteryLimited,
+  applyFirmwareRange,
   BATTERY_SAFE_MAX_WATTS,
 } from "./devices";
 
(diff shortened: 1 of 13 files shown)


srsholmes:
Closing — this branch has been split, and everything shippable from it now lives elsewhere.

- #254 — Wi-Fi 7 `iwlmld` recovery (merged/open on its own)
- #255 — Intel `i915`/`xe` GPU temperature sensors
- #256 — Intel handheld TDP + GPU control, with AMD paths deliberately untouched (draft, awaiting hardware testing)

What was only ever in this PR is the remaining idea: letting generic `firmware-attributes` discovery outrank `ryzenadj` on **every** AMD device, rather than only on Intel as #256 ships it. That changes detection order and the write unit on ROG Ally / Legion Go hardware we have no way to test, and #256 already carries a cheaper safety net for the same failure — a rejected write is retried in the other unit, which rescues a mis-scaled rail without any precedence change.

So the value left here is a generic path for *future* AMD handhelds, not a fix for a known-broken one. Not worth an untested reordering on shipped hardware, so this is closed rather than parked.

If it becomes worth revisiting, the evidence needed is on a ROG Ally or Legion Go with `asus-armoury` / `lenovo-wmi-other-N` bound:

1. `cat /sys/class/firmware-attributes/*/attributes/ppt_pl1_spl/max_value` — settles the watts-vs-milliwatts question outright.
2. Whether the TDP slider works at all on `main` today.
3. Whether #256's retry-in-the-other-unit path recovers it.

The branch is left in place, so this can be reopened rather than rebuilt.

Issue #263, opened 2026-08-21
Overlay's X11 Steam-window lookup matches a localized WM_NAME, so non-English clients fall to the scored heuristic


srsholmes:
Same root cause as #259, on the X11 side instead of CDP. Split out of #262, which fixed the two CDP call sites.

## The bug

`findSteamWindow()` in `apps/loadout-overlay/src/bun/native/gamescope-atoms.ts:277` takes its strongest-signal path from an exact string compare:

```ts
if (name === "Steam Big Picture Mode") {
```

That string is the value of Steam's `SP_WindowTitle_BigPicture` localization key, so the window is named differently in every non-English client:

| locale | WM_NAME |
| --- | --- |
| english | `Steam Big Picture Mode` |
| russian | `Режим Big Picture` |
| german | `Big-Picture-Modus` |
| schinese | `Steam 大屏幕模式` |
| finnish | `Steamin televisiotila` |
| turkish | `Steam Geniş Ekran Modu` |

(Read from `~/.local/share/Steam/steamui/localization/steamui_*-json.js` — all 32 languages differ, and no ASCII substring is common to them.)

So on a non-English client step 1 never matches and we always drop to step 2, the scored heuristic. Per the comment at `gamescope-atoms.ts:251-254` and `docs/overlay-gamescope-integration.md:212-215`, that is precisely the path that "frequently latched onto the wrong steamwebhelper child — so the reclaim watcher was monitoring a window gamescope ignores". Live xprop surveys on an OXP Apex found ~10 windows of class `steamwebhelper`/`steam` competing.

The scored fallback may well pick correctly much of the time — `STEAM_GAME = 769` plus `_NET_WM_WINDOW_TYPE_NORMAL` is a decent filter. But every non-English user is permanently running on the path we wrote *because* it was unreliable, and step 1 exists to make the choice unambiguous.

## Why it wasn't fixed in #262

The CDP fix had a locale-independent replacement to hand: `browserType=4`, Steam's own marker, verified against a live Gaming Mode capture. There's no equivalent property established for the X11 window, and window detection can't be verified from Desktop Mode — it needs a gamescope session to test against, ideally a non-English one.

## Likely shape of a fix

[...]

PR #264, opened 2026-08-21
fix(overlay): prefer the window that has STEAM_OVERLAY when picking Steam's BPM

srsholmes:
Refs #263. Branched off `main`, independent of #262.

## The bug

`_pickBpmWindow`'s tiebreak can't distinguish an **absent** property from one **present and set to 0**:

```ts
if ((atoms.get("STEAM_OVERLAY") ?? 0) !== 0 || (atoms.get("STEAM_INPUT_FOCUS") ?? 0) !== 0)
```

A live survey on a Deck in Gaming Mode ([details on #263](https://github.com/srsholmes/loadout/issues/263)) shows the structural filter leaves a pool of **two** windows, not one:

```
0x3000035  1920x1200  STEAM_GAME=769  STEAM_OVERLAY=0  TYPE=NORMAL  "Steam Big Picture Mode"  ← real BPM
0x2800003   200x200   STEAM_GAME=769  (absent)         TYPE=NORMAL  "steamwebhelper"          ← renderer helper
```

The existing comment already anticipates the helper — "STEAM_GAME=769 alone also matches the 200×200 renderer helper" — but `_NET_WM_WINDOW_TYPE` doesn't separate them, since both are NORMAL. In BPM home the real window sits at `STEAM_OVERLAY=0`, so nothing reads as "asserting", and the pool falls through to `pool[0]` — which candidate order makes the helper. The recorded trace agrees:

```
findSteamWindow → 0x2800003 (scored, candidates: 0x2800001,0x2800003,0x1e00001,
                             0x1c00005,0x1e00018,0x1c0000c,0x1c0000e,0x1000041,0x3000035)
```

English clients never hit this: tier 1 matches `WM_NAME == "Steam Big Picture Mode"` and returns before the scoring runs. On a non-English client that title is Steam's translated `SP_WindowTitle_BigPicture` ("Режим Big Picture"), tier 1 can't fire, and every lookup lands on the mis-picked window.

## The fix

One extra tier between "asserting" and `pool[0]`: prefer a window that *has* `STEAM_OVERLAY` at all. Steam sets it on its own BPM window as self-state and only there — verified as Steam's write rather than ours, since the overlay had touched no window in that Steam session. `_hasAtom()` already existed.

## Why this is deliberately small

This path is hard to test across devices, controllers and displays, so the change is bounded so it can't regress hosts I can't reach:

- **It only reorders within the existing pool.** It cannot shrink the candidate set and adds no new "no window found" path.
- **With no candidate carrying the property, the result is byte-identical to today** — there's a test that strips `STEAM_OVERLAY` from both windows and asserts we still return `pool[0]`.
- **The English fast path is untouched** — a test asserts tier 1 still short-circuits.
- **No geometry / largest-window heuristic.** It would cover the residual case where Steam hasn't set the property yet on a fresh boot, but a fullscreen-size assumption is exactly what breaks on external displays and unusual modes. Falling back to today's behaviour there is the safer trade.

A trace line records when the new tier fires, so a field report can tell the tiers apart:

```
[gamescope-atoms] _pickBpmWindow → 0x3000035 (has STEAM_OVERLAY; pool: 0x2800003,0x3000035)
```

## Testing

Three tests modelling the surveyed window set (localized title, English fast path, and the no-property fallback). Full backend suite (3231), typecheck and lint clean.

Also run against the live window set on the Deck, rather than only against mocks:

```
pool (NORMAL + STEAM_GAME=769): 0x2800003 0x3000035
tier A: asserting?      0x2800003 → none      0x3000035 → 0,0
tier B (new): present?  0x2800003 → absent    0x3000035 → present
old behaviour: pool[0] = 0x2800003
new behaviour:           0x3000035
real BPM (WM_NAME) =     0x3000035
```

## What this does and doesn't change for users

The everyday open/close path never depended on this. `show()` writes the atoms that make the overlay appear onto **our own** window; resolving Steam's window only matters for clearing Steam's `STEAM_OVERLAY` when it is asserting `1`, and for the `hide()` write that prevents the input halt described at `docs/overlay-gamescope-integration.md:117-121`. That's why the overlay tests fine on a Russian client today. This makes those two edge cases correct rather than fixing a visible everyday failure.

### DIFF
--- a/apps/loadout-overlay/src/bun/native/gamescope-atoms.test.ts
+++ b/apps/loadout-overlay/src/bun/native/gamescope-atoms.test.ts
@@ -880,4 +880,115 @@ HDMI-1 connected primary 1920x1080+0+0 (normal left inverted right x axis y axis
       expect(xdotoolWindowVerbs()).toEqual([]);
     });
   });
+
+  // Issue #263. Reproduces the window set surveyed live on a Deck in Gaming
+  // Mode: two windows pass the STEAM_GAME=769 + _NET_WM_WINDOW_TYPE filter,
+  // neither asserts overlay/focus, and only the real BPM window carries
+  // STEAM_OVERLAY at all (reading 0). The renderer helper comes first in
+  // candidate order, so `pool[0]` used to hand back the wrong window.
+  describe("findSteamWindow — BPM vs renderer helper (#263)", () => {
+    const HELPER = "0x2800003"; // 200x200 steamwebhelper, no STEAM_OVERLAY
+    const BPM = "0x3000035"; // 1920x1200 real BPM, STEAM_OVERLAY=0
+
+    /** Route xprop/xdotool for the surveyed window set. `bpmTitle` stands in
+     *  for the localized WM_NAME so tier 1 never fires. */
+    function mockSurveyedWindows(bpmTitle: string) {
+      mockRun.mockImplementation((cmd: string[]) => {
+        if (cmd.includes("xdotool")) {
+          // Both windows come back from the class search, helper first.
+          if (cmd.includes("--class")) {
+            return Promise.resolve({
+              stdout: `${parseInt(HELPER, 16)}\n${parseInt(BPM, 16)}\n`,
+              exitCode: 0,
+            });
+          }
+          return Promise.resolve({ stdout: "10\n", exitCode: 0 });
+        }
+        if (!cmd.includes("xprop")) {
+          return Promise.resolve({ stdout: "", exitCode: 0 });
+        }
+        const idIdx = cmd.indexOf("-id");
+        const id = idIdx >= 0 ? cmd[idIdx + 1] ?? "" : "";
+        const isBpm = Number(id) === parseInt(BPM, 16);
+        const asks = (a: string) => cmd.includes(a);
+
+        if (asks("_NET_WM_NAME") || asks("WM_NAME")) {
+          const title = isBpm ? bpmTitle : "steamwebhelper";
+          return Promise.resolve({
+            stdout: `_NET_WM_NAME(UTF8_STRING) = "${title}"\n`,
+            exitCode: 0,
+          });
+        }
+        // Both are managed, both tagged with Steam's own appID.
+        if (asks("_NET_WM_WINDOW_TYPE")) {
+          return Promise.resolve({
+            stdout: "_NET_WM_WINDOW_TYPE(ATOM) = _NET_WM_WINDOW_TYPE_NORMAL\n",
+            exitCode: 0,
+          });
+        }
+        if (asks("STEAM_GAME")) {
+          return Promise.resolve({
+            stdout: "STEAM_GAME(CARDINAL) = 769\n",
+            exitCode: 0,
+          });
+        }
+        // The crux: present-but-zero on BPM, absent on the helper.
+        if (asks("STEAM_OVERLAY")) {
+          const overlay = isBpm
+            ? "STEAM_OVERLAY(CARDINAL) = 0"
+            : "STEAM_OVERLAY:  not found.";
+          const focus = asks("STEAM_INPUT_FOCUS")
+            ? isBpm
+              ? "\nSTEAM_INPUT_FOCUS(CARDINAL) = 0"
+              : "\nSTEAM_INPUT_FOCUS:  not found."
+            : "";
+          return Promise.resolve({ stdout: `${overlay}${focus}\n`, exitCode: 0 });
+        }
+        return Promise.resolve({ stdout: "", exitCode: 0 });
+      });
+    }
+
+    it("picks the BPM window, not the helper, when the title is localized", async () => {
+      mockSurveyedWindows("Режим Big Picture");
+      const atoms = new GamescopeAtoms({
+        display: ":0",
+        windowName: "X",
+        forceXprop: true,
+      });
+      expect(await atoms.findSteamWindow()).toBe(BPM);
+    });
+
+    it("still takes the WM_NAME fast path on an English client", async () => {
+      mockSurveyedWindows("Steam Big Picture Mode");
+      const atoms = new GamescopeAtoms({
+        display: ":0",
+        windowName: "X",
+        forceXprop: true,
+      });
+      expect(await atoms.findSteamWindow()).toBe(BPM);
+    });
+
+    it("falls back to the previous answer when no candidate has STEAM_OVERLAY", async () => {
+      // Guarantees the change can only reorder within the pool: strip the
+      // property from both and we land on pool[0], exactly as before.
+      mockSurveyedWindows("Режим Big Picture");
+      const prev = mockRun.getMockImplementation();
+      mockRun.mockImplementation((cmd: string[]) => {
+        if (cmd.includes("xprop") && cmd.includes("STEAM_OVERLAY")) {
+          return Promise.resolve({
+            stdout: "STEAM_OVERLAY:  not found.\n",
+            exitCode: 0,
+          });
+        }
+        return prev!(cmd);
+      });
+      const atoms = new GamescopeAtoms({
+        display: ":0",
+        windowName: "X",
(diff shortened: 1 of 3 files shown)


srsholmes:
## Russian verification on device — the new tier fires and picks correctly

Steam in Russian, Gaming Mode, #264 installed:

```
10:41:42Z _pickBpmWindow  → 0x3000035 (has STEAM_OVERLAY; pool: 0x2800003,0x3000035)
10:41:42Z findSteamWindow → 0x3000035 (scored, candidates: 0x2800001,0x2800003,0x1e00001,
                                       0x1c00005,0x1e00018,0x1c0000c,0x1c0000e,0x1000041,0x3000035)
```

The pool is exactly the two windows from the survey, and `0x3000035` is the window titled `Режим Big Picture`. Candidate order puts `0x2800003` first, so on `main` this same call returns the 200×200 renderer helper. Both halves are now confirmed end-to-end on hardware: English takes `(by WM_NAME)` and never reaches the new tier, Russian reaches it and lands on the real window.

### Testing note: the resolution is cached, so open/close won't re-exercise it

Worth recording for anyone testing this later. `steamWindowId` is cached — `findSteamWindow` only re-runs when it is null, or when the reclaim watcher sees ~2s of quiet atoms. Opening and closing the overlay repeatedly after a language switch reuses whatever was resolved earlier, so it looks fine while touching none of this code.

Switching Steam's language restarts *Steam* but not the *overlay*, so the overlay kept an English-era `(by WM_NAME)` resolution across the switch. Forcing a fresh resolution needs an overlay restart — and it should use the graceful stop-wait-pkill-start sequence from `scripts/install-local.sh:243-267`, since a plain `systemctl --user restart loadout-overlay` races CEF's profile lock and can return a blank webview.

### Incidental: the cached id survives a Steam restart

The overlay held `0x3000035` across Steam restarting for the language change, and afterwards that id was still the live BPM window — X11 reusing ids deterministically rather than anything guaranteeing correctness. Nothing misbehaved and this PR doesn't change it, but a Steam restart could in principle leave the overlay holding a stale id until the reclaim watcher re-resolves. Worth remembering if an "overlay stopped taking focus after Steam restarted" report ever appears.

srsholmes:
Review follow-ups applied in 84e69b5. All five findings were real; two of them undercut the safety claim in the PR description above.

**1. The tier ran on the unfiltered pool.** `pool` falls back to the raw class-search result when nothing passes the `STEAM_GAME=769` filter — menu/tooltip popups, VRStream, 10×10 utilities. Promoting one of those over `pool[0]` would have changed the desktop-Steam and unsurveyed shapes the fallback exists to preserve, which is precisely the shape that can't be tested here. Now gated on `managed.length > 0`, so "reorders within a filtered pool" actually holds. New test covers it.

**2. `hide()` could manufacture the signal the tier reads.** Unlike `show()`'s zero-pass (guarded by `_steamSnapshotIsAsserted()`), the restore write was unconditional, stamping `STEAM_OVERLAY` onto whatever window we had resolved. After a mis-pick, one open/close cycle marked the renderer helper; the helper then satisfied the presence tier for the rest of the session, and the trace printed `(has STEAM_OVERLAY; …)` as though it had worked. Our own bookkeeping became the evidence for repeating the mistake — and the fix could never recover once Steam did set the property on the real window.

Now skipped when the show()-time snapshot carried no `STEAM_OVERLAY`. Safe because with the atom absent gamescope never treated the window as an overlay candidate, so there is no claim to release and the input-halt reasoning at `docs/overlay-gamescope-integration.md:117-121` doesn't apply. Both reclaim-path callers of `_zeroSteamFocusAtoms` are already gated on Steam asserting `=1`, so they can't create it either. New test asserts no `STEAM_OVERLAY` write reaches Steam's window in the fresh-boot shape, and that our own window still tears down normally.

**3. Redundant reads, and an xprop-only dependency.** Both `_readAtoms` paths already drop absent atoms (`x11.ts:277` only `set()`s on a non-zero-length reply; the xprop regex never matches `not found.`), so the asserting loop already had the presence answer. `_hasAtom` was spawning an extra xprop per candidate on every cold resolve and quiet-tick re-resolve — and being xprop-only, it would have silently returned false on a host running fine over libxcb without xprop, so the tier would never have fired there. Presence is now recorded during the existing read.

**4. The "English fast path" test was vacuous.** It gave only the BPM window `STEAM_OVERLAY`, so the new tier returned the same window and deleting tier 1 wouldn't have failed it. The helper now carries the property too and sorts first, so returning BPM can only mean tier 1 short-circuited. Confirmed by disabling tier 1 and watching the test fail, then restoring.

**5. Stale docs.** The tier list in `docs/overlay-gamescope-integration.md` now includes the presence tier and its filtered-pool scoping, plus a note that step 1's `WM_NAME` is a localized string.

34 tests in the file, full backend suite (3250), typecheck and lint clean. Not re-verified on device since these changes — the on-hardware runs recorded above were against 507df83.

srsholmes:
## Final verification on device, against the reviewed code (84e69b5)

All four behaviours observed in the live trace on a Steam Deck in Gaming Mode, not just in tests.

**English — tier 1 still short-circuits, new tier never runs:**

```
14:12:00Z findSteamWindow → 0x3000035 (by WM_NAME) candidates=9
```

```
$ xprop -id 0x3000035 WM_NAME STEAM_GAME STEAM_OVERLAY
WM_NAME(UTF-8) = "Steam Big Picture Mode"
STEAM_GAME(CARDINAL) = 769
STEAM_OVERLAY(CARDINAL) = 0        # 1920x1200
```

**Russian — reaches the presence tier and picks the real window:**

```
14:19:01Z _pickBpmWindow  → 0x3000035 (has STEAM_OVERLAY; pool: 0x2800003,0x3000035)
14:19:01Z findSteamWindow → 0x3000035 (scored, candidates: 0x2800001,0x2800003,0x1e00001,
                                       0x1c00005,0x1e00018,0x1c0000c,0x1c0000e,0x1000041,0x3000035)
```

`WM_NAME` on that window reads `Режим Big Picture`, so tier 1 cannot match. Candidate order puts `0x2800003` first, so `main` returns the 200×200 helper here.

**The unfiltered-pool guard (review finding 2) — declined to promote, matching `main`.** Caught by accident when the overlay happened to start before Steam:

```
14:06:50Z findSteamWindow → 0x1e00001 (scored, candidates: 0x1e00001,0x1c00005)
```

Neither candidate carries `STEAM_GAME`, so `managed` was empty and `pool` was the raw class-search list — exactly the shape the gate protects. `0x1e00001` is a 10×10 window titled "steam"; the tier correctly stayed out of it and returned `pool[0]`.

**The `hide()` no-create guard (review finding 1) — fired for real, on that same mis-resolved window:**

```
14:08:20Z hide: Steam's window carried no STEAM_OVERLAY at show() — leaving it absent
          rather than creating one
```

That is the poisoning path being blocked in the wild. Before the fix we would have stamped `STEAM_OVERLAY=0` onto a 10×10 window, and it would then have satisfied the presence tier for the rest of the session. Conversely, on the real BPM window the guard correctly does *not* suppress the write — every hide there logs `wrote STEAM_OVERLAY=0 (snap was 0)`, so the input-halt protection still runs.

## Not covered

- **`hide()` with a game running.** `STEAM_GAMES_RUNNING` logged `count=unset` on every cycle, matching the "Steam writes inconsistently" note in `docs/overlay-gamescope-integration.md`, so the `gameAlive=true` restore branch may be hard to reach at all. Unchanged by this PR either way.
- **Desktop Steam / non-Deck shapes.** Untested, which is why the new tier is gated to a filtered pool — on those shapes it returns `pool[0]` exactly as today.

## Follow-ups, not for this PR

- #263 stays open: this fixes the fallback's *choice*, but tier 1 still matches a localized `WM_NAME`, so non-English clients keep taking the fallback rather than the fast path.
- The startup race is now observed rather than inferred — an overlay that starts before Steam resolves a junk window and caches it, with no invalidation for "Steam wasn't running when I resolved". Worth tracking separately.

PR #266, opened 2026-08-21
fix(fan-control): persist the global fan mode across restarts

### LINKED ISSUE #265: Global fan preset (Silent/Balanced/Performance) does not persist across reboots
### What happened?

In fan-control plugin, when selecting a fan curve preset (e.g. "Silent" or "Balanced") or custom fan curve, the dynamic fan curve works during the session. However, after rebooting
  the device or restarting `loadout.service`, the active preset is lost. The fan defaults to a fixed static manual speed instead of resuming the selected curve
  preset.

Root Cause
    1. `FanControlBackend.applyPreset()` in `plugins/fan-control/backend.ts` sets the in-memory `this.activePreset` and starts the curve loop, but does not
  persist the selection to plugin storage (`mutatePluginStorage`).
    2. In `FanControlBackend.onLoad()`, only per-game profiles and custom curve point definitions are loaded; no global preset or active curve mode is restored or
  re-applied.
    3. On reboot, `activePreset` is `null` and no `curveInterval` loop is running. `getFanInfo()` falls back to reading sysfs `pwm1_enable` (which drivers like
  `oxp_ec` initialize to `1` / manual) and the current PWM value, causing the UI to display a static manual fan setting.

Proposed Solution
    - Persist the selected mode / active preset (`activePreset: PresetName | "custom" | "auto" | "manual"`) in plugin storage when changed (`applyPreset`,
  `applyCustomCurve`, `setFanMode`, `setFanSpeed`).
    - In `onLoad()`, restore the saved preset/mode and invoke `applyPreset()` or `applyCustomCurve()` once hardware detection completes.

### Device

OneXPlayer APEX

### OS / distro

Bazzite 44

### Loadout version / commit

0.8.2

### Logs

```shell

```

### Is this specific to a plugin?

Yes — a specific plugin (name it above)

srsholmes:
Fixes #265.

## The bug

Reported on an OXP APEX / Bazzite: pick "Silent", reboot, and the fan is back on a fixed manual duty.

`applyPreset` set `activePreset` in memory and started the curve loop but wrote nothing to plugin storage, and `onLoad` restored only per-game profiles and the custom curve's *points*. After a restart nothing was running the curve, and `getFanInfo` fell back to reading `pwm1_enable` — which drivers like `oxp_ec` initialise to `1` (manual) — so the UI showed a static manual speed.

The reporter's root-cause analysis was exact and I verified each part against the code before writing anything. This is their proposed fix.

## What it does

The user's global choice now persists as a tagged union and is restored on load:

```ts
type GlobalFanMode =
  | { kind: "preset"; name: PresetName }
  | { kind: "custom" }
  | { kind: "manual"; percent: number | null }
  | { kind: "auto" };
```

Four details worth review attention:

**Restored from the hardware scanner's `onFound`, not `onLoad`.** Applying a preset means writing PWM, and on hosts where the driver module (oxpec and friends) lands after our service starts there is nothing to write to yet — the scanner already retries every 30s for exactly that case. Guarded with a flag so a later rescan can't stomp a selection the user has changed since the first restore.

**Per-game profiles must not write it.** The `PerGameEngine` drives the same `setFanSpeed`/`setFanMode` entry points, so persisting unconditionally would promote a game's fan profile to the system default the next time the plugin loaded — and exiting the game wouldn't undo it, because the restore path goes through the same methods. All four engine call sites now pass `persist: false`, as does the restore itself.

**The requested speed is saved, not the safety-floor-raised one.** `setFanSpeed` applies the #97 safety floor, which can raise the duty well above what the user asked for. Persisting that would bake a moment at 95 °C into the saved preference and restore it on the next boot.

**Storage is user-editable JSON**, so the value is sanitised on read exactly as `sanitiseCurve` does for the curve: an unknown preset name restores nothing rather than throwing during `onLoad`, and a hand-typed `percent: 900` is clamped before it can reach a PWM write.

An install that has never chosen a mode keeps the previous behaviour of leaving the fans to whatever the firmware set.

## Structural note

`setFanSpeed`'s body moved to a private `applyUserFanSpeed` so the public entry can persist on success — note this is *not* the existing `setFanSpeedInternal`, which is the bare duty write the safety watchdog and curve loop use, and which must keep bypassing all of this. Same shape for `setFanMode` → `applyUserFanMode`.

## Testing

Full backend suite (3264), typecheck and lint clean. 14 new tests: the sanitiser in `lib/global-mode.test.ts`, and behaviour in `backend.test.ts` covering persistence of each mode, restore-on-hardware-found, restore-once, the malformed-value path, and the safety-floor distinction.

The per-game test drives the real engine through `handleGameLaunch`/`handleGameExit` rather than passing `persist: false` by hand — mutation-checked by reverting the two engine call sites, which fails the test.

**Not yet verified on device.** This is plugin-only, so it deploys via `prepare-plugins.sh` without a binary rebuild.

### DIFF
--- a/plugins/fan-control/app.spec.tsx
+++ b/plugins/fan-control/app.spec.tsx
@@ -168,6 +168,47 @@ describe("fan-control plugin", () => {
     });
   });
 
+  it("tracks the live duty in the slider while a preset drives the fan", async () => {
+    // A preset puts the hardware in manual mode but the *curve* owns the
+    // duty, rewriting it every 2s. The slider used to keep showing the
+    // user's last manual value (or its 50/100 default), so it disagreed
+    // with the "Fan speed" row right above it.
+    callMock.mockImplementation((method: string) => {
+      if (method === "getFanInfo")
+        return Promise.resolve({
+          ...mockFanInfo,
+          mode: "manual" as const,
+          activePreset: "silent",
+          fans: [{ index: 0, rpm: 1500, pwm: 77, percent: 30 }],
+        });
+      if (method === "getCustomCurve") return Promise.resolve([]);
+      return Promise.resolve(null);
+    });
+    const container = createContainer();
+    const { mount } = await import("./app");
+    mount(container);
+
+    await waitFor(() => {
+      const slider = container.querySelector('input[type="range"]');
+      expect(slider).not.toBeNull();
+      expect(Number((slider as HTMLInputElement).value)).toBe(30);
+    });
+
+    // And it follows the curve as the duty moves.
+    const handler = eventHandlers.get("fan-update");
+    expect(handler).toBeDefined();
+    handler?.({
+      ...mockFanInfo,
+      mode: "manual" as const,
+      activePreset: "silent",
+      fans: [{ index: 0, rpm: 2600, pwm: 140, percent: 55 }],
+    });
+    await waitFor(() => {
+      const slider = container.querySelector('input[type="range"]');
+      expect(Number((slider as HTMLInputElement).value)).toBe(55);
+    });
+  });
+
   it("shows unavailable message when no fan hardware", async () => {
     callMock.mockImplementation((method: string) => {
       if (method === "getFanInfo")
--- a/plugins/fan-control/app.tsx
+++ b/plugins/fan-control/app.tsx
@@ -198,6 +198,15 @@ function FanControl() {
         setCustomActive(true);
         setActivePreset(null);
       }
+      // While a preset or the custom curve is driving, the slider is a
+      // readout, not a setting: the curve rewrites the duty every 2s and
+      // the user's last manual value has nothing to do with what the fan
+      // is doing. Track the live percent so it agrees with the "Fan speed"
+      // row above it (issue #265 follow-up).
+      const live = info.fans?.[0]?.percent;
+      if ((info.activePreset || info.customCurveActive) && typeof live === "number") {
+        setSliderValue(live);
+      }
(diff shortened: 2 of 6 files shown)


srsholmes:
Superseded by #267, which is the same feature rebuilt as a single commit off `main`.

This branch reached six commits and 1119 insertions for what is really two sentences of behaviour — remember the user's fan choice, and show the duty the fan is actually running at. Most of the growth was machinery defending against races introduced while building it: an intent-epoch guard with a `supersededSince()` check required after *every* await, commit-point flag sequencing, and two extra fields tracking what the user had chosen.

Every problem that machinery solved is real, and this PR's review history is the record of how each was found — worth keeping rather than squashing away:

- a per-game profile could become the user's global default
- a superseded apply could leave `activePreset` set with no curve loop, which **mutes the safety watchdog** (it defers its write when that flag is set)
- `handleGameExit`/`handleGameLaunch` interleaving on a game switch left the curve driving the fan while a profile was meant to own it — caught on device, not in review
- `setFanMode("manual")` laundering a game's duty, or a safety-floored 100%, into the saved preference

#267 keeps all of those invariants but replaces the per-await obligation with a single operation lock, the same pattern already used in `plugins/disable-controller-input`. The rule is one line — public entry points serialize, private methods assume the lock is held — instead of an analysis every future edit has to re-derive. Two fields collapse into one, the snapshot change disappears, and the `persist` flag becomes a public/private split that also makes lock re-entry impossible by construction.

Closing unmerged; nothing here is lost.

PR #269, opened 2026-08-24
feat(apex): global rumble intensity for OneXPlayer handhelds

### LINKED ISSUE #260: 7.2 kernel (oxp-hid)
7.2 kernel was released today. Any plans on making a rumble control plugin later? I heard this kernel includes new oxp-hid driver that's actually supporting such kind of manipulation now (i.e. for Onexplayer Apex console). 
It may also fix battery charge limit/passthrough mode stuff for Apex.

srsholmes:
Adds a **Vibration** card to the OneXPlayer plugin: the gamepad's global rumble level (`rumble_intensity`, 0–5), a single attenuator the MCU applies to everything — so it scales rumble even in games that ignore Steam's own setting. Closes #260.

## Why it lives in the OneXPlayer plugin, not its own

This started as a standalone plugin. It isn't one, because the capability isn't general. `rumble_intensity` is a `hid-oxp` attribute, and surveying Handheld Daemon's device tree, no other handheld exposes an equivalent through the kernel:

| Vendor | Global rumble level? | Mechanism |
| --- | --- | --- |
| OneXPlayer | yes, 0–5 | `hid-oxp` sysfs attribute |
| GPD Win | yes, off/medium/high | `wincontrols` vendor HID config blob — no sysfs |
| MSI Claw | no | force-feedback events only (a buzz, not a level) |
| ROG Ally | no | force-feedback events only |
| Ayaneo / Legion Go / Orange Pi | no | — |

The shell gives every loaded plugin a sidebar row whether or not its hardware is present, so a standalone plugin was a permanently dead "No rumble control found" entry on every Deck, Ally and Legion. The card now appears only where the attribute exists, and hides itself on gen-1 OneXPlayers where the driver exposes RGB only.

Storage moves from the `vibration` key to `apex.rumbleIntensity`. That strands nothing — this never shipped, so nobody has a stored level, which made now the only free moment to do it.

## Detection is by attribute, not device

Globs `/sys/bus/hid/devices/*/rumble_intensity` rather than matching a VID/PID table. The attribute exists on every gen-2 device the driver binds — X1 mini series, G1 A/i, AOKZOE A1X, Apex, X2 Mini Pro — so a table would be a list to maintain and a way to exclude working hardware. A retry scan covers `hid-oxp` binding after we load.

## The stored value wins over the driver's

`rumble_intensity` reads back a driver-side cache initialised to maximum in `oxp_cfg_probe`; it never queries the MCU. So it reports 5 after every module load regardless of what the firmware is doing. We persist the user's choice, re-apply it on load, and label a driver-sourced value as unreliable in the UI.

## Tests

13 unit tests for the control, 8 for the pure range/label helpers, 5 UI. Mutation-checked: preferring the driver value over the stored one, restoring with nothing stored, dropping the clamp, failing the call on a storage error, showing the card without hardware, hardcoding 0–5, and a dead click each fail a test.

Sysfs is injected. The original read the host's real `/sys`, which made the plugin's own backend tests behave differently on a machine with OneXPlayer hardware attached — precisely the machine they get run on.

## Verified on hardware

Detection, level changes and persistence exercised on a OneXPlayer Apex.


srsholmes:
Superseded by #274, which contains these commits plus the fingerprint self-heal. It's all the same plugin and easier to review and test as one branch.

Issue #232, opened 2026-07-28
Battery charge limit is a no-op on OneXPlayer APEX — EC ignores charge_control_end_threshold


srsholmes:
## Summary

The charge limit does nothing on the OneXPlayer APEX. The plugin writes `charge_control_end_threshold` correctly and sysfs reads it back correctly — the EC/firmware simply ignores it and keeps charging.

Observed live with the limit set to **80%**, battery climbing straight past it.

## Evidence

Plugin state is correct end-to-end:

```
~/.config/loadout/plugins/battery-tracker.json
{ "chargeLimitPercent": 80, "bypassMode": "disabled" }

[battery-tracker] battery candidates: BATT(15), hid-…-battery(-140) — picked BATT
[battery-tracker] Using battery at /sys/class/power_supply/BATT
[battery-tracker] charge control: limit=true, bypass=charge_behaviour, awake=true
[battery-tracker] restored charge limit 80%
```

sysfs confirms the write landed:

```
/sys/class/power_supply/BATT/charge_control_end_threshold  → 80
/sys/class/power_supply/BATT/charge_behaviour             → [auto] inhibit-charge inhibit-charge-awake
/sys/class/power_supply/ACAD/online                       → 1
```

And the battery charges past it anyway — sampled 20 s apart:

```
t+00s  capacity=89  status=Charging  end_threshold=80
t+20s  capacity=89  status=Charging  end_threshold=80
t+40s  capacity=90  status=Charging  end_threshold=80
```

Reported at 87%, 90% by the time sampling finished. No plugin errors in the journal.

## Why it can't be worked around in the plugin (as things stand)

- The attribute is provided by the **generic ACPI battery driver** (`/sys/bus/acpi/drivers/battery`), not by a vendor driver. The kernel exposes it because ACPI advertises charge control; the EC doesn't honour it.
- **There is no alternative path.** `charge_control_end_threshold` on `BATT` is the *only* `charge_control*` node in all of sysfs. The OneXPlayer EC driver `oxpec` is loaded but exposes only `tt_toggle` and hwmon — no charge control.
- There is no `charge_control_start_threshold`, so the usual "only enforced when the cycle starts below the threshold" mitigation isn't available either.
- **The failure is undetectable from sysfs**: the read-back returns the value we wrote, so the plugin cannot tell success from a no-op and will always report the limit as applied.

## Relationship to #224

[...]

PR #270, opened 2026-08-25
feat(doctor): report device hardware, so "does my handheld work" is answerable

srsholmes:
Prompted by a **OneXPlayer X2 Mini Pro** report: fan control says "no hardware", and the SoC reached 84 °C at 80 W before the user shut the device down.

## Why

`loadout-doctor.sh` reads no DMI, no hwmon, and no module list. So for the most common new-device report — "fan control finds nothing" — its output cannot distinguish a Loadout bug from a kernel driver that never bound, which is almost always the real cause.

For that user specifically: `oxpec` provides the `oxp_ec` hwmon node and binds by DMI alias. It ships 34 aliases — `ONEXPLAYERAPEX`, `ONEXPLAYERX1MiniPro`, `ONEXPLAYERG1A` … — and **no X2 entry**. No binding, no `fan1_input`, nothing to write to. The current doctor output would never have shown that.

## What section 10 adds

- **DMI** — `sys_vendor`, `product_name`, `product_version`, `board_name`, `bios_version`. The exact `product_name` decides whether a driver's DMI table matches, which TDP profile `matchDevice()` picks, and what an upstream patch would need to add.
- **Every hwmon node** with which of `fanN_input` / `pwmN` / `pwmN_enable` / `fanN_target` it carries — the same predicates `scanHwmonDevices()` uses, so the output says directly whether fan control *could* work.
- **Loaded handheld platform/HID drivers**, plus whether `/etc/modprobe.d/hid-oxp.conf` is still on disk (it disables rumble, gamepad mode and button remapping — see #268).
- **`ectool hello`**, not `command -v ectool`. The binary exists on hardware where the handshake fails, so the current optional-tools "yes" is misleading; `hello` is what fan control actually probes with.
- **HID rumble / gamepad-mode attributes**, current temperatures, and the backend's own plugin detection lines (section 9 greps the overlay journal only).

## Output on a working device

Run on an Apex — the healthy case, for contrast with what the X2 will show:

```
  --- DMI ---
  sys_vendor:      ONE-NETBOOK
  product_name:    ONEXPLAYER APEX

  --- hwmon (fan control needs fanN_input + pwmN + pwmN_enable, or a writable fanN_target) ---
  hwmon5 k10temp                temp1_input
  hwmon6 oxp_ec                 fan1_input pwm1 pwm1_enable
  hwmon9 amdgpu                 temp1_input

  --- platform / HID drivers ---
  loaded: hid_oxp oxpec

  --- ectool ---
  ectool: not installed

  --- HID device attributes ---
  0003:1A86:FE00.0005    rumble_intensity=5 rumble_intensity_range=0-5 gamepad_mode=xinput
```

On the X2 Mini Pro the `oxp_ec` line and the `oxpec` module will both be absent — diagnosis in one paste.

## Notes

POSIX `sh`, read-only, no sudo, matching the existing sections' style — the script is safe to ask any user to pipe from `curl`. `sh -n` clean, and run end to end on an Apex.

Shipping this before any device-support code deliberately: it turns future reports into evidence rather than hearsay, which is the habit `CLAUDE.md`'s triage section already argues for — and it's what would have caught the #232 misdiagnosis.

### DIFF
--- a/scripts/loadout-doctor.sh
+++ b/scripts/loadout-doctor.sh
@@ -331,6 +331,45 @@ done
 [ "$found_hid" -eq 1 ] || echo "  no HID device exposes rumble/gamepad-mode attributes"
 
 echo ""
+echo "  --- USB devices ---"
+# The full table, not just a probe for ids we already know: on unfamiliar
+# hardware the useful question is what this device's internal gamepad and
+# fingerprint reader enumerate AS, which is what would let us extend the
+# id lists the OneXPlayer plugin matches on.
+if command -v lsusb >/dev/null 2>&1; then
+    lsusb 2>/dev/null | sed 's/^/  /'
+    echo ""
+    echo "  known ids the OneXPlayer plugin looks for:"
+    for pair in "1a86:fe00 gamepad HID MCU" "045e:028e gamepad (X360-compatible)" "2808:c652 fingerprint reader"; do
+        id=${pair%% *}
+        label=${pair#* }
+        if lsusb -d "$id" >/dev/null 2>&1; then
+            printf "    %-12s %-32s present\n" "$id" "$label"
+        else
+            printf "    %-12s %-32s ABSENT\n" "$id" "$label"
+        fi
+    done
+else
+    echo "  lsusb not installed — cannot check the gamepad or fingerprint reader"
+fi
+
+echo ""
+echo "  --- xHCI controllers ---"
+# The gamepad-recovery fix rebinds one of these. Which one is normally read
+# out of the kernel log; the bound list is what it would act on.
+if [ -d /sys/bus/pci/drivers/xhci_hcd ]; then
+    for l in /sys/bus/pci/drivers/xhci_hcd/0000:*; do
+        [ -e "$l" ] || continue
+        echo "  bound: $(basename "$l")"
+    done
+else
+    echo "  no xhci_hcd driver directory"
+fi
+echo "  --- controller died in this boot's log? ---"
+dmesg 2>/dev/null | grep -iE "xhci_hcd .*(HC died|assume dead)" | tail -5 | sed 's/^/  /' \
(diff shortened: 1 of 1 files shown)


srsholmes:
Extended in c8eb82f, because the first cut didn't actually answer the question it needs to.

Section 10 covered the fan side (DMI, hwmon, platform drivers) but reported **nothing** the OneXPlayer plugin's own features depend on — so it couldn't tell us whether un-gating that plugin (#271) will work on hardware we don't have:

| #271 detects with | reported before | now |
| --- | --- | --- |
| `isOneXPlayer()` — DMI | ✅ | ✅ |
| TDP row — `product_name` | ✅ | ✅ |
| `lsusb 2808:c652` — fingerprint reader | ❌ | ✅ |
| `lsusb 1a86:fe00` / `045e:028e` — gamepad | ❌ | ✅ |
| `dmesg` "HC died" — which controller to rebind | ❌ | ✅ |
| bound `xhci_hcd` PCI node | ❌ | ✅ |

The USB table is dumped **in full**, not just probed for the ids we already know. On unfamiliar hardware the useful question isn't "are our ids present" — it's what that device's gamepad and reader enumerate *as*, which is what would let us extend `GAMEPAD_IDS`. The known-id check sits alongside as a one-line verdict.

On this Apex:

```
  known ids the OneXPlayer plugin looks for:
    1a86:fe00    gamepad HID MCU                  present
    045e:028e    gamepad (X360-compatible)        present
    2808:c652    fingerprint reader               present

  --- xHCI controllers ---
  bound: 0000:65:00.4
```

That `0000:65:00.4` is the hardcoded fallback the recovery uses, so the same block on an X2 Mini Pro says immediately whether the gamepad fix can work there — and if the ids differ, exactly what to add.

`sh -n` clean, still read-only, still no sudo. `dmesg` may need root on some distros; that line degrades to a note rather than failing.

PR #273, opened 2026-08-26
feat(input-plumber): supply a device config when InputPlumber has none

srsholmes:
Follow-up to #271. Separate PR because it's a different plugin and a different risk class: #271's claims are all backed by a doctor report, whereas a device config is a *behavioural* change on hardware we can't test.

## The problem

InputPlumber only builds a `CompositeDevice` for hardware one of its configs matches, on the exact `/sys/class/dmi/id/product_name` (no shipped OneXPlayer config uses a wildcard). A handheld it doesn't recognise gets **no composite device at all** — `listCompositeDevicePaths()` returns empty, the wake-button picker lists nothing, overlay intercept has nothing to act on. The whole plugin looks broken on hardware that's fine.

That's the OneXPlayer X2 Mini Pro as of InputPlumber 0.78.1: a stock `045e:028e` XInput pad and the same `1a86:fe00` MCU the Apex and X1 configs already use, and nothing naming it.

## Why a drop-in works

`get_devices_paths()` reads `/etc/inputplumber/devices.d` **before** `/usr/share/inputplumber/devices`, with ties broken by basename. So the config uses the filename upstream would use — it *shadows* a future upstream config rather than loading alongside it and matching the same machine twice. When upstream ships one, the UI says so and suggests removing ours, since theirs is tested on the hardware.

(Note for later: the capability-map search order is **inverted** — `/usr/share` first — so the same trick doesn't work there.)

## What's deliberately left out

Only enough to make a composite device exist. Not included:

- **The `AT Translated Set 2 keyboard` source.** The Apex config claims it for volume keys. Claiming the system keyboard on a board we've never run means a wrong guess costs the user their keyboard, not a feature.
- **A back-paddle capability map.** On the X2 the MCU's intercept mode no longer carries the paddles; HHD remaps them to keycodes so they arrive as KEY_F15/KEY_F16. But that mapping is applied by *HHD's* init and we'd be relying on the kernel `hid-oxp` to do the same (untested), and F16 is already the overlay's wake key while F15 is its capture sentinel. Claiming them would change behaviour that may already work — that needs a decision and an `evtest`, not a default.

## Safety

The card only renders when we have a config for this exact machine **and** InputPlumber has produced no composite devices. A device that already works is never offered one. DMI matching is exact on both fields — deliberately stricter than `packages/devices`' substring matching, because a false positive here hands input handling to a config written for different hardware.

Install and remove both restart InputPlumber; removal treats "already gone" as success.

## Tests

17 unit tests + 7 UI tests. Every config is parsed with `Bun.YAML` and checked for the shape InputPlumber requires, and the DMI in the YAML body is cross-checked against the metadata (they're two hand-written copies of one fact). Mutation-checked: substring matching, a Loadout-branded filename, a prettified DMI string, adding the AT keyboard, and dropping either UI guard each fail a test.

## Not verified

We have no X2 Mini Pro. This is the reviewable form of the question rather than a tested change — worth getting in front of the user who reported it before it ships. If it works on hardware, this is also the basis for an upstream InputPlumber PR (their ONEXPLAYER 3 support, #660, is the same three-file shape).

### DIFF
--- a/plugins/input-plumber/README.md
+++ b/plugins/input-plumber/README.md
@@ -4,6 +4,37 @@
 
 Installs the InputPlumber input-routing daemon that other controller features rely on, and quietly does nothing if it's already present. Mostly a one-time setup helper so the rest 'just works'.
 
+## Device configs
+
+InputPlumber only builds a `CompositeDevice` for hardware one of its configs
+matches, and matching is on the exact `/sys/class/dmi/id/product_name`. On a
+handheld it has no config for there is no composite device at all — so the
+wake-button picker lists nothing and overlay input intercept has nothing to
+act on. Every InputPlumber-dependent feature is inert on hardware that is
+otherwise completely conventional.
+
+For devices we can describe, this plugin offers to drop a config into
+`/etc/inputplumber/devices.d`. That directory is searched *before*
+`/usr/share/inputplumber/devices`, with ties broken by basename — so these
+files deliberately use the name upstream would use, and *shadow* a future
+upstream config rather than loading alongside it and matching the same
+machine twice. When upstream does ship one, the UI says so and suggests
+removing ours, since theirs is tested on the hardware.
+
+The card only appears when we have a config for this exact machine **and**
+InputPlumber has genuinely produced no composite devices. A device that
+already works is never offered one.
+
+Currently shipped:
+
+| Device | DMI `product_name` | Notes |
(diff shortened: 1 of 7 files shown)


srsholmes:
Closing in favour of input plumber dealing with adding support for devices properly to prevent conflicts / maintaining the separation of concerns.

PR #275, opened 2026-08-28
fix(apex): the wake block counts as applied on distros we don't stage a karg on

srsholmes:
On an **Apex running CachyOS or Bazzite**, `applied` was permanently false even though the PME layer was genuinely blocking the wake:

```
cachyos: applied=false  kargActive=false  kargApplicable=true
bazzite: applied=false  kargActive=false  kargApplicable=true
steamos: applied=true
```

`kargApplicable` conflated *"we know this board's GPIO pin"* with *"we can actually put the karg there"*. Off SteamOS we deliberately never edit the bootloader, so `kargActive` could only become true if the user added it by hand.

## Why it matters more now

- **The toggle springs back to off.** The user enables the block, PME *is* applied, and the control flips back — whose natural response is to flip it again, calling `revert()` and undoing the one path that worked. This is the same defect fixed in #274 for *unmeasured* boards; measured boards on a non-SteamOS distro were still exposed.
- **The self-heal fires every boot.** `shouldHeal` keys on `!applied`, so on those distros it would re-apply and toast *"restored after a system update"* on every single boot, forever. That ships in #274, so this wants to land before any release carrying it.

## Fix

`kargAutomatic` — a confirmed board **and** a bootloader we edit — now drives `applied`. `kargApplicable` keeps its meaning (the pin is known) and still gates the manual hint shown off SteamOS.

So per distro, for a measured board:

| | PME layer | GPIO karg | `applied` |
| --- | --- | --- | --- |
| SteamOS | automatic | automatic | needs both |
| CachyOS / Bazzite / Arch | automatic | manual hint only | PME alone |
| Unmeasured board, any distro | automatic | withheld | PME alone |

## Tests

Six new, covering all three distro classes plus the unmeasured board, and asserting the heal stays quiet where PME alone is the whole fix while still firing when the block genuinely goes missing. Mutation-checked: reverting to `kargApplicable`, and forcing `kargAutomatic` to either constant, each fail multiple tests.

Found by checking whether the fix actually worked on CachyOS and Bazzite rather than assuming. Two adversarial review passes missed it because every existing test pinned `distro: "steamos"`.

### DIFF
--- a/plugins/apex/lib/fingerprint.test.ts
+++ b/plugins/apex/lib/fingerprint.test.ts
@@ -597,3 +597,71 @@ describe("shouldHeal", () => {
     );
   });
 });
+
+describe("an Apex on a distro whose bootloader we don't edit", () => {
+  /**
+   * CachyOS/Bazzite: the board is measured (so we know the pin) but we never
+   * stage the karg there. Counting it toward `applied` anyway left the toggle
+   * springing back to off with PME genuinely blocking the wake — and made the
+   * startup self-heal re-apply and toast "restored after a system update" on
+   * every boot, forever.
+   */
+  const pmeApplied = {
+    [UDEV_RULE_PATH]: "(rule)",
+    [`/sys/bus/pci/devices/${CTRL}/power/wakeup`]: "disabled",
+  };
+
+  it.each([["cachyos"], ["bazzite"], ["arch"]])(
+    "%s: PME blocking alone counts as applied",
+    async (distro) => {
+      const { deps } = makeFpDeps({
+        files: { ...pmeApplied },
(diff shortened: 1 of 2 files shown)


srsholmes:
Closing — the premise is wrong. Making `applied` true with only the PME path closed would have the UI report the wake block as on while the GPIO path is still live, so a touch could still wake the device. The repo says so in three places (fingerprint.ts:7 "both must close", the origin script's success condition, and #162's commit message).

The springing-toggle bug is real, but the honest fix is to bind the toggle to the user's stored intent rather than to `applied`, and to actually stage the karg on CachyOS and Bazzite. Folding into that work.

